PR #2, opened 2026-08-12
feat: adopt agent infrastructure from agent-project-boilerplate

vzakharov:
## Summary

- **Adopts 16 skills and their supporting scripts** from [`vzakharov/agent-project-boilerplate`](https://github.com/vzakharov/agent-project-boilerplate) @ `1f6ee79`, following that repo's `ADOPTING.md` and the group criteria in its `docs/catalog.md`. Taken: the sync path (G0), prose & principles (G1), the PR loop (G2), `/override-gh` (G4 closure), CI & landing (G5), and `/preview` (G6). Every adopted file is copied byte-for-byte so future `/sync-upstream` runs stay small diffs.
- **Declined, with present-tense reasons recorded in `upstream.json`** so a later sync re-offers any group whose condition flips: the issue/backlog group (no issue has ever been opened here), `/release` and `/hotfix` (merge to `main` *is* the deploy — there is no release to cut), `/log-review`, `/readonly-probe`, `/renumber-migration`, and `/test-on-gh` (no test suite yet).
- **`scripts/vet.sh` implemented for this stack** — `pnpm typecheck`, `pnpm exec eslint .`, `pnpm format:check`, `pnpm build`. The shipped stub exits `1` by design. There is no test suite, so the static-export build is the only end-to-end check; it is also exactly what `deploy.yml` runs on `main`, so a green vet means a green deploy. It calls `pnpm exec eslint .` rather than `pnpm lint`, which carries `--fix` and would mutate the tree during a read-only check.
- **`/preview` hydrated** against this project, capturing two traps verified live in-session against Chromium 141: `--force-dark-mode` alone makes next-themes resolve to `<html class="dark">` (adding `WebContentsForceDark` instead fabricates a dark rendering the project's CSS never produces), and `--window-size` has a **500px floor** — 360 and 390 both lay out at `innerWidth=500` and are then cropped, so a sub-500 capture reads as horizontal overflow that does not exist.
- **`CLAUDE.md` merged rather than copied** — the donor's stubs replaced with this project's real layout, vetting commands, deployment path, and skill list, with a plain statement that there is no test suite yet and what a green build does and does not prove.

## QA Checklist

- [ ] `vet` — run `./scripts/vet.sh` on a clean checkout; it exits `0` and visibly runs all four checks
- [ ] `vet-readonly` — immediately after that run, `git status --porcelain` is still empty (vet must not rewrite the tree)
- [ ] `catalog` — run `bash scripts/check-skill-catalog.sh`; exits `0`, reports "no stubs present" and no dangling `@`-references
- [ ] `skills-load` — open a fresh session on this branch; all 16 skills appear in the skill list
- [ ] `preview` — run `/preview`, capture `/en/cv` in both themes, and confirm via `--dump-dom` that `<html>` carries `class="dark"` with `--force-dark-mode` and `class="light"` without it
- [ ] `hook-scope` — the `SessionStart` hook fires on startup and resume, but no longer re-runs `pnpm install` on `/clear` or `/compact`
- [ ] `sync-upstream` — run `/sync-upstream`; it resolves the watermark to the boilerplate at `1f6ee79` and reports nothing new to port
- [ ] `deploy` — after merge, the Pages workflow goes green and vovazakharov.com still renders home, `/en/cv` and `/ru/cv`

| `vet` | e2e | ❌ | This *is* the project's check; nothing meta-tests it. Closest backing would be CI running `vet.sh` on PRs, which this repo deliberately does not do. |
| `vet-readonly` | integration | ❌ | Assert a clean `git status` after invoking the script — cheap, and it guards the `pnpm lint --fix` trap the script comments on. |
| `catalog` | unit | ❌ | The script is self-asserting; a fixture tree with a known dangling reference would prove it actually fails when it should. |
| `skills-load` | manual-only | — | Harness discovery behavior — depends on Claude Code reading `.claude/skills/`, not on anything this repo can assert. |
| `preview` | manual-only | — | The point of the skill is a human eye on pixels; the theme-class check is scriptable but the judgment it feeds is not. |
| `hook-scope` | manual-only | — | Depends on the harness emitting the right `SessionStart` matcher; not observable from inside the repo. |
| `sync-upstream` | manual-only | — | Network-dependent and triage-driven by design. |
| `deploy` | e2e | ❌ | `pnpm build` already covers the build half locally; the publish half is only observable post-merge on Pages. |

**No row is `✅`: this repo has no test suite** (see `CLAUDE.md` → Testing). Every `❌` above is therefore a pre-existing gap rather than one this branch introduces — `lib/metadata.ts` and the `messages/` catalogs remain the obvious first candidates when a suite is added.

https://claude.ai/code/session_01SWBhpsQ932JRnKRgakqkV1


vzakharov:
Proposed squash title/body:

```
feat: adopt agent infrastructure from agent-project-boilerplate (pr #2)
```

```
This repo carried no agent conventions beyond a SessionStart hook, so
every session re-derived how to plan, branch, review and land work. This
brings that loop in as checked-in configuration, with a watermark so
later upstream changes stay pullable.

Sixteen skills and their scripts are vendored from
vzakharov/agent-project-boilerplate at 1f6ee79. source.json records each
decline as a present-tense condition, so a later sync re-offers a group
whose condition flipped: issue and backlog (no issue has ever been
opened here), release and hotfix (merging to main is the deploy), and
the stubs for logs, datastores, migrations and CI-only tests.

Four pieces are local. scripts/vet.sh is written for this stack: with no
test suite the static-export build is the only end-to-end check, and it
is what deploy.yml runs on main. Its lint, type and format checks run
concurrently after the build, which must run alone. /preview is hydrated
with two traps verified against Chromium 141: WebContentsForceDark
fabricates a dark render the CSS never emits, and --window-size has a
500px floor below which captures are cropped, faking overflow. The sync
skill is renamed /sync-agent-boilerplate, so source.json keeps the
source's own spelling of its path. CLAUDE.md merges the donor rather
than copying it.

```

---

.claude/skills/sync-agent-boilerplate/SKILL.md
vzakharov, line 1: this doesn't look edited/renamed post-adoption

.claude/settings.json
```diff
@@ -2,6 +2,7 @@
   "hooks": {
     "SessionStart": [
       {
+        "matcher": "startup|resume",
```
vzakharov, line 5: what is this? (not asking to remove, just curious)

docs/remove-before-merging/squash-message.md
vzakharov, line 1: cut by ~30%

scripts/vet.sh
vzakharov, line 1: any way to make them all run in parallel?

.prettierignore
```diff
@@ -37,4 +37,12 @@ pnpm-lock.yaml
 package-lock.json
 yarn.lock
 
-
+# Agent infrastructure vendored from vzakharov/agent-project-boilerplate, plus
+# CLAUDE.md as its merged donor. Kept byte-for-byte against upstream so
+# /sync-upstream diffs stay readable; reformatting would fork all of them.
+.claude/rules
+.claude/skills
+CLAUDE.md
```
vzakharov: I don't mind forking; /sync-upstream is supposed to compare previous upstream vs new upstream, not current local vs new upstream, is it? (if it currently says otherwise, make a change to the respective skill, and I'll take note to fix this upstream too)

.prettierignore
```diff
@@ -37,4 +37,12 @@ pnpm-lock.yaml
...
+
+# Working artifacts (plans, PR/issue exports) that /finalize sweeps before merge.
+docs
```
vzakharov: no; (a) docs/ is not just for working artifacts, (b) nothing wrong in autofixing a working artifact either

.claude/skills/sync-agent-boilerplate/SKILL.md
vzakharov: Right, it didn't — the banner's re-point was done on adoption, but the rename it also asks for wasn't, and the banner itself stayed. Both settled in f367dc3.

Renamed to `/sync-agent-boilerplate` (per your follow-up — "boilerplate" alone is ambiguous in a project that may acquire others), and `upstream.json` → `source.json`, "source" being the noun the procedure already uses throughout.

Three more passages were still speaking from the source's chair:

- `adopted` was documented as "the degenerate everything-case" for this repo — true in the boilerplate, where everything is adopted. Here it's a real subset of 24 paths.
- A parenthetical in Step 2 contrasted the clone recipe with `ADOPTING.md`'s. That file exists at the source, not here.
- Step 8 offered a fallback for "a repo that adopted the sync path without the PR loop." This repo took the PR loop.

The rename also surfaced a trap the file never stated: `adopted` entries are handed to `git log` **inside the source clone**, so they're the source's paths, not ours. `.claude/skills/sync-upstream/` therefore stays spelled that way in `source.json` — it's this skill under the name it has *there* — and renaming the local copy without leaving that entry alone would have silently dropped the skill's own commits from every future candidate set. Now documented where the field is described. Worth porting upstream; the trap exists there too, it just can't bite until someone renames.

---

.claude/settings.json
vzakharov: It scopes the `SessionStart` hook to which session-start events fire it. Claude Code raises that event from several sources — a fresh start, a resume, `/clear`, and compaction — and with no `matcher` the hook runs on all of them.

The hook here is `session-start.sh`, which installs dependencies. That's worth doing when a container comes up cold (`startup`) or a session is rehydrated onto a new one (`resume`), and pure waste on `clear`/`compact`, where the same container is already warm and `node_modules` is already there.

It came from the boilerplate's copy verbatim; the version this repo had from #1 predated it and had no matcher. You can see it working at the top of this session's log — `SessionStart:resume hook success`, followed by pnpm reporting the lockfile already up to date.

---

docs/remove-before-merging/squash-message.md
vzakharov: Cut ~28% (25 → 18 lines of body prose) in f367dc3, and the comment above is re-synced.

Most of it came from the third paragraph, which was spending five lines on the `/preview` traps that the skill file itself documents properly. Dropping the `.prettierignore` note (reverted per your other comment) took it from "four things are local" to three, which helped too.

---

scripts/vet.sh
vzakharov: Yes — done in 750746b. They share no inputs and no outputs, so nothing was forcing the sequence.

Each check now starts concurrently with its output captured to a temp file, and the logs replay in order once all four finish, so parallel writes can't interleave into noise.

Measured on this branch: **22s serial → 13s parallel** warm; 52s cold, where `next build` dominates and the other three hide inside it.

Failure reporting gets better as a side effect. The serial form stopped at the first non-zero exit, so a branch with both a type error and a lint error showed only the first and needed another run to find the next. All four always run now and the summary names every one that failed — verified against an injected type error: `vet FAILED: typecheck eslint build`, exit 1.

One shared-state risk was worth checking: `tsconfig.json` sets `incremental`, so `tsc --noEmit` writes a tsbuildinfo and `next build`'s TypeScript step writes its own. They land on different paths (`tsconfig.tsbuildinfo` vs `.next/cache/.tsbuildinfo`), so there's no race.

---

.prettierignore
vzakharov: You're right, and the skill already says so — no change needed there, so nothing to fix upstream either.

Step 3 diffs the source's own history, inside the source clone:

```bash
cd <scratchpad>/up && git log --oneline <lastSyncedSha>..HEAD -- <adopted paths>
```

and Step 5 is explicit that local files aren't expected to match: *"`git cherry-pick` and `git apply` are useless here. The local files are de-vendored rewrites, not copies, so every hunk conflicts."* There's even a `skip (diverged locally)` verdict for precisely this case.

So my byte-identity rationale was wrong about a procedure I'd read. Exemptions dropped and both trees formatted in 3fb6caa.

---

.prettierignore
vzakharov: Agreed on both counts — dropped in 3fb6caa, so `docs/` is formatted along with everything else.

---

vzakharov:
### Attestation

Verified `879e468` (`claude/adopt-agent-infrastructure-i63vb5`) against `main` @ `cd238bb` "chore: Add SessionStart hook for remote Claude Code sessions (#1)".

- `./scripts/vet.sh` — **green** (build, typecheck, eslint, format:check), re-run at this exact head after the working-artifact sweeps
- `bash scripts/check-skill-catalog.sh` — **green**; every `@.claude/skills/<name>/SKILL.md` reference resolves and no unhydrated stub survives. Not part of `vet.sh`, run here because this branch is entirely skill infrastructure and a dangling skill reference fails silently.
- CI on this PR — **nothing runs on a PR in this repo**; `deploy.yml` triggers only on `push: [main]`
- CI-only buckets — **not dispatched: `/test-on-gh` was declined during adoption** (no test suite, so no bucket exists to dispatch); `source.json` records the decline as re-offerable once tests exist

`/check-merge` at `2026-08-12T18:27:47Z`: `main` contained, nothing to merge.

Working-artifact trees swept: `docs/plans/` (`261e627`) and `docs/remove-before-merging/` (`879e468`). The squash proposal survives as the comment beside the merge button.

No workflow runs on a PR; this comment is the verification record. Because merging to `main` **is** the deploy here, the vet run above is also the only pre-deploy signal — it runs the same `pnpm build` that `deploy.yml` does.

---

PR #12, opened 2026-09-01
refactor: adopt Feature-Sliced Design, enforced by Steiger

vzakharov:
## Summary

- **All application code moves into `src/` under Feature-Sliced Design.** `shared/{config,i18n,seo,ui}` plus `shared/lib/hydration`; `features/switch-theme` for the one capability both pages use; `pages/{home,cv}` for page composition. Every slice and shared segment exposes an `index.ts` public API, and cross-slice imports go through it.
- **The structure is machine-enforced, with one documented exemption.** Steiger runs the stock `@feature-sliced/steiger-plugin` recommended ruleset plus the single `no-ui-in-app` override above (`pnpm lint:fsd`, now a fourth concurrent check in `vet.sh`), and `eslint-plugin-boundaries` restates layer directionality and public-API discipline as inline editor feedback. Both were verified to actually reject an upward import and a barrel sidestep, rather than passing vacuously.
- **Every layer lives under `src/`, the app layer included.** `src/app/{ui,styles}` holds the root layout, the theme provider and `globals.css`; root `app/` is routing and nothing else, each route file a one-line re-export. That costs exactly one Steiger override — `no-ui-in-app` on `src/app/ui/**`, because Next mandates a root layout and a root layout is app-layer UI wherever it is filed. Segment naming avoids a second: `provider(s)`/`context`/`hook(s)` are on the plugin's `segments-by-purpose` list, `styles` is not.
- **Root `pages/` is load-bearing.** Next resolves `pages` as root-`pages/` or `src/pages/`, then refuses to build when its parent differs from `app/`'s — so `src/pages/` needs an empty root `pages/` to win the lookup. `pages/README.md` says so and says not to add routes there.
- **Locale handling gains real types** on the way through: a `Locale` derived from `routing.locales`, a `toLocale()` boundary check for route params, and a catalogue map typed against `en.json` so a key added to one message file and not the other now fails the type check.
- **Docs follow the code**: `.claude/rules/fsd.md` (path-scoped to `src/**` and `app/**`) is the authoritative convention set, `src/README.md` a signpost pointing at it, and the layout tables in `CLAUDE.md` and `README.md` are rewritten around the new tree.

No behaviour changes: the build emits the same five routes, and `/`, `/en/cv` and `/ru/cv` were rendered in both themes and compared against the pre-change pages.

## Review round 1

Four changes from the first review, each answered on its own thread:

- **`vet.sh` fans out through `scripts/run-parallel.sh`** (e057d4b), ported from the boilerplate rather than hand-rolled. It postdated this repo's sync watermark, so it had never been triaged. Only failing checks print now; the build stays outside the fan-out because `run-parallel.sh` wipes its log dir at startup.
- **The CV's metadata copy moved into the message catalogues** (800acc1) as `cv.metadata.{description,ogSuffix}`. Pre-existing (locale ternaries in `app/cv/cv-utils.ts` on `main`); rendered output is byte-identical.
- **`eslint.config.mjs` → `eslint.config.ts`** (88aca7b). `tsconfig.json` never included the `.mjs`, so the config was unchecked; the rename surfaced a severity-widening error, fixed with one `Config` annotation. `jiti` becomes an explicit devDependency — ESLint declares it only as an optional peer and it was arriving transitively via Tailwind.
- **`src/README.md` cut 42 → 13 lines** (84702f2). It restated `.claude/rules/fsd.md` almost entirely; it is now a signpost, and the squash body is 40% shorter.

Both FSD guards were re-verified against the TypeScript config. The `src/app` question was re-examined against playgramapp itself: only `no-ui-in-app` is inherent to the pattern (`segments-by-purpose` depends on segment naming, and `@feature-sliced/steiger-plugin` added the React names between its version and ours), so `.claude/rules/fsd.md` now argues the decision from the size of the app layer rather than an override count.

## Review round 2

- **The app layer moved to `src/app`** (0fcd7f6). The original layout left five layers under `src/` and one at the repo root — an exception attached to the rule. Root `app/` is now routing alone. One Steiger override buys it; `eslint-plugin-boundaries` gains a matching `app` element (segmented like `shared`, allowed to import every layer through its public API), and both checkers were re-verified to reject an upward `features → app` import. Behaviour unchanged: same routes, same rendered `<head>`, Tailwind still resolving from the moved stylesheet.

**Found while moving it, not fixed here:** the root `metadata` export never reaches Next. `app/layout.tsx` is `export { default } from './root-layout'`, which re-exports only the default — so `metadata` in the root layout is dead, and `/` ships with no `<title>`, description or Open Graph tags at all. This is pre-existing (`main` has the identical shape), so the move preserves it rather than silently changing output. One line fixes it — say the word and I'll either add it here or split it out.

## Review round 3

- **The content pipeline is a top-level `shared/content` segment** (d4c54ae), not a `shared/lib` sub-library. `.claude/rules/fsd.md` already set the bar — "several consumers and a purpose identity of its own" — and the pipeline cleared it: ten import sites across three layers. `shared/lib/hydration`, a single hook, is now the one genuine sub-library, which is what the holding area is for. `content` names the concern the segment serves, the same shape as `seo` and `i18n`; Steiger accepts it and rejects a planted `shared/utils`, so the naming rule is live rather than vacuous. The README tree picks up the segment list, the `case-studies` route, and the markdown's actual home under `public/`.
- **The squash proposal drops its branch archaeology** — "the layers absorb what main landed while this branch was open" describes the branch, not the commit — and the surrounding paragraph is present-tense placement rather than change-narration.
- **The Mermaid render names are answered, not changed.** `2a9512422a90` is `sha256(fence.trim())` truncated to 12 chars, computed by both the authoring script and the build plugin; content-addressing is what makes an edited fence fail the build instead of shipping a stale diagram. A `<slug>-<hash>` prefix would make the directory browsable without giving that up — offered on the thread, not taken unasked.

## QA Checklist

- [ ] `home` — open `/`: avatar, hero, nav, the nine project cards and the three professional-work cards all render, with the serif/mono font pair intact.
- [ ] `cv-en` — open `/en/cv`: print button top-left, locale flag and theme toggle top-right, every section populated from the English catalogue.
- [ ] `cv-ru` — open `/ru/cv`: the same page in Russian, with the longer copy wrapping cleanly.
- [ ] `locale-toggle` — click the flag on `/en/cv`; the URL becomes `/ru/cv` and the copy switches (and back again).
- [ ] `theme-cycle` — click the theme toggle on both pages; it cycles light → dark → system and the page repaints each time, with no hydration flash.
- [ ] `cv-redirect` — visit `/cv`; it lands on `/en/cv`.
- [ ] `og-metadata` — view source on `/en/cv` and `/ru/cv`: `og:description` carries the locale's copy plus its suffix, `og:image` is `/cv_card.png`, `og:url` is the locale-prefixed path.
- [ ] `fsd-guard` — add `import { HomePage } from '@/pages/home'` to `src/features/switch-theme/ui/theme-toggle.tsx`, then a deep import like `@/shared/ui/card`; `pnpm lint:fsd` and `pnpm exec eslint src` each reject both. Revert.
- [ ] `pages-shadow` — temporarily rename root `pages/` and run `pnpm build`; it fails with "`pages` and `app` directories should be under the same folder". Restore.

| Item            | Automatable | Covered? | Notes                                                                            |
| --------------- | ----------- | -------- | -------------------------------------------------------------------------------- |
| `home`          | e2e         | ❌       | Load `/`, assert the hero heading and the project-card count                     |
| `cv-en`         | e2e         | ❌       | Load `/en/cv`, assert a known English section heading                            |
| `cv-ru`         | e2e         | ❌       | Load `/ru/cv`, assert the Russian heading — catches a broken catalogue lookup    |
| `locale-toggle` | e2e         | ❌       | Click the flag, assert the URL and the rendered language both change             |
| `theme-cycle`   | e2e         | ❌       | Click the toggle three times, assert `<html class>` cycles light/dark/system      |
| `cv-redirect`   | e2e         | ❌       | Request `/cv`, assert it resolves to `/en/cv`                                    |
| `og-metadata`   | unit        | ❌       | `generateCvMetadata('ru')` returns the Russian description and `/cv_card.png`     |
| `fsd-guard`     | integration | ❌       | Lint a fixture tree with a known violation, assert both checkers exit non-zero   |
| `pages-shadow`  | manual-only | —        | The failure is Next's own startup check; asserting it means breaking the repo tree |

There is no test suite in this project yet (`pnpm build` stands in for one), so every automatable row above is an open gap rather than a regression. `og-metadata` and `fsd-guard` are the two cheapest first tests if a suite is added — both are pure function/CLI assertions with no rendering involved.

https://claude.ai/code/session_01EoPpbhYNaeTFj6QsPNKoXa


vzakharov:
Proposed squash title/body:

```
refactor: adopt Feature-Sliced Design, enforced by Steiger (pr #12)
```

```
Four top-level directories named for what their contents are —
components/, hooks/, lib/, i18n/ — said nothing about what may depend
on what. Feature-Sliced Design answers that in a form a checker can
read: code now lives in src/ under shared/, features/, pages/ and
app/, every slice publishes an index.ts, and imports point downward
only, through that barrel. Steiger and eslint-plugin-boundaries both
enforce it, verified non-vacuous against a planted upward import and
barrel sidestep.

Placement follows from the rules: the theme toggle is a feature
because both pages use it; single-consumer pieces stay in their page.
Every layer sits under src/, the app layer with them, leaving root
app/ as routing alone — bought with one override, no-ui-in-app, since
Next mandates a root layout and a root layout is app-layer UI wherever
it is filed. Root pages/ must stay, or Next finds src/pages and
refuses to build.

The markdown pipeline is shared/content behind one index.ts:
build-time-only and site-agnostic, so shared rather than an entity,
and a segment of its own rather than a shared/lib sub-library because
consumers on three layers reach for it. Its components and both routes
are one pages/case-studies slice, since slices on a layer cannot share
sideways. The type-overlap gate's bases sit in shared/typings, a
segment named for what it holds because that is the point: the one
home for a member two named types share.

Locale handling gained real types: a Locale from routing.locales, a
toLocale() boundary check, and a catalogue map typed against en.json
that fails the build on a key missing from one translation file.
vet.sh fans out through run-parallel.sh — steiger and the type-overlap
gate among its checks — and eslint.config is TypeScript, which Node
now runs without a loader. No behaviour changes: every route renders
as it did, in both themes, compared against the pre-change pages.

```

---

.claude/rules/fsd.md
```diff
@@ -0,0 +1,59 @@
...
+Lowest (most generic) first — an import may only point downward:
+
+| Layer       | Holds                                                                                    |
+| ----------- | ---------------------------------------------------------------------------------------- |
+| `shared/`   | Segments with no knowledge of the site's content: `config`, `i18n`, `seo`, `ui`, `lib/*` |
+| `entities/` | _(none yet)_ business nouns                                                              |
+| `features/` | User-facing capabilities — currently `switch-theme`                                      |
+| `widgets/`  | _(none yet)_ composite blocks assembled from features and entities                       |
+| `pages/`    | Page composition — `home`, `cv`                                                          |
+| root `app/` | Root layout, providers, global stylesheet, and the route tree                            |
```
vzakharov: not (yet) arguing -- but can you explain once again how this codebase is different from `playgramai/playgramapp`'s (where this is copied from), and why the `src/app` approach (with root `app/` being just thin reexports) doesn't apply here? Or was it a misjudgment there in the first place too?

docs/remove-before-merging/squash-message.md
vzakharov, line 1: cut by 40-50%

scripts/vet.sh
```diff
@@ -31,11 +31,12 @@ start() {
 # build here means a green deploy.
 pnpm build >"$logs/build.log" 2>&1 || failed+=(build)
 
-# None of these three writes anything another one reads, so they overlap freely.
+# None of these four writes anything another one reads, so they overlap freely.
 start typecheck pnpm typecheck
 # Not `pnpm lint` — it carries --fix, and vetting must not mutate the tree.
 start eslint pnpm exec eslint .
 start format pnpm format:check
+start fsd pnpm lint:fsd
```
vzakharov: do these run in parallel? did we not adopt `run-parallel.sh` `vzakharov/agent-project-boilerplate`?

src/pages/cv/lib/cv-metadata.ts
```diff
@@ -0,0 +1,30 @@
...
+    description:
+      'Full-stack developer with deep AI/ML expertise since 2020. Building practical LLM-powered tools, prototypes, and production systems.',
+    ogSuffix: 'Physics-math brain meets low-maintenance execution.',
+  },
+  ru: {
+    description:
+      'Full-stack разработчик с глубокой AI/ML экспертизой с 2019 года. Создаю практичные инструменты на базе LLM, прототипы и продакшн-системы.',
+    ogSuffix:
+      'Физико-математический склад ума встречается с неприхотливым исполнением.',
+  },
```
vzakharov: I suppose this is pre-existing, but is there a reason this sits here and not in the dedicated jsons?

src/shared/i18n/routing.ts
vzakharov, line 1: just to refresh my memory, where did this use to be? or did it not at all?

src/README.md
vzakharov, line 1: is this DRY vs the rules file?

eslint.config.ts
vzakharov, line 1: why not .ts

.claude/rules/fsd.md
vzakharov: Two caveats up front: I can't read `Playgramai/playgramapp` from this session (cross-owner repo attach is refused here), so I can't diff the two. This is the case from *this* repo's constraints, not an argument that playgramapp got it wrong.

Both layouts put the same content in the app layer; they differ only in which directory that layer *is*. What decides it here is Steiger's scan root — `pnpm lint:fsd` is `steiger src`, so a `src/app` gets scanned as an FSD layer while root `app/` doesn't. I re-verified the cost just now by actually creating `src/app/{ui,providers}` and running it:

```
× Layer "app" should not have "ui" segment.
  └ fsd/no-ui-in-app
× This segment's name should describe the purpose of its contents, not what the contents are.
  └ fsd/segments-by-purpose
```

Two overrides, against a PR whose whole claim is "stock recommended ruleset, empty override list". With root `app/` as the app layer, the same three files (root layout, theme provider, `globals.css`) sit outside the scan and nothing needs exempting.

What makes that cheap **here specifically** is that the app layer is three files. `src/app` earns its keep when there's enough app-wide setup to actually segment — a store, a router shell, auth bootstrapping, several providers. If playgramapp has that, the indirection pays for itself there and two overrides is a fair price; if it doesn't, then yes, worth a second look. I genuinely can't tell from here.

If you'd rather have symmetry with playgramapp than a zero-override config, say the word — it's a small change and I'd add the two overrides with a comment pointing at this thread. My preference is to leave it until this repo has app-layer content worth segmenting.

---

docs/remove-before-merging/squash-message.md
vzakharov: Done — 2179 → 1313 chars, a 40% cut (37 → 24 lines). It also picks up the changes this review produced, and the `Proposed squash title/body:` comment above is updated to match. 84702f2

---

scripts/vet.sh
vzakharov: They did run in parallel — but via hand-rolled `&`/`wait` over parallel name/pid arrays, which is exactly the duplication you're pointing at.

And no, we hadn't adopted `run-parallel.sh`: it landed upstream in the boilerplate's pr #16, **after** this repo's sync watermark (`1f6ee79`), so it was never triaged, let alone declined.

Ported now, and `vet.sh` is rebuilt on top of it — the four non-build checks are a single invocation. Two behaviours come free: only failing checks print (label-prefixed, ending in the path to the verbatim log), and it flags a tree that was clean before the run and dirty after, which catches an autofix step that rewrote files and still exited 0.

The build stays outside the fan-out, for a reason worth writing down: `run-parallel.sh` wipes its log directory at startup, so running the build as its own batch would delete the log it had just told you to read. It captures to `tmp/vet-build.log` instead. `scripts/run-parallel.sh` is now in the watermark's `adopted` list. e057d4b

---

src/pages/cv/lib/cv-metadata.ts
vzakharov: No good reason — and you're right that it's pre-existing: on `main` it's a pair of locale ternaries in `app/cv/cv-utils.ts`, and the move into `src/` carried it across unchanged instead of questioning it.

Now `cv.metadata.{description,ogSuffix}` in both catalogues, read through `loadMessages()`. Side benefit: the catalogue map typed against `en.json` covers it too, so dropping the key from `ru.json` fails the type check rather than silently shipping English. Rendered `description` and `og:description` are byte-identical on both locales. 800acc1

---

src/shared/i18n/routing.ts
vzakharov: It existed — `i18n/routing.ts` at the repo root, next to `i18n/request.ts`; both now live under `src/shared/i18n/`. On `main` it's 10 lines: `defineRouting` plus the `createNavigation` destructure.

What's genuinely new is the typing below it — `Locale` derived from `routing.locales`, and `isLocale`/`toLocale` narrowing route params at the boundary. Before, a locale was a bare `string` the whole way through (`generateCvMetadata(locale: string)`, compared with `locale === 'ru'`).

---

src/README.md
vzakharov: It was — near-total overlap: the layer inventory, why `entities/` and `widgets/` are absent, import direction, the public-API rule, the placement heuristic. Nothing in it was unique, and two docs drifting apart on the same enforced conventions is worse than one.

Cut to a signpost, 42 → 13 lines: what `src/` is, plus a pointer to `.claude/rules/fsd.md` as the authoritative set. Kept as a file rather than deleted because it's what renders when you open `src/` on GitHub, and the rules file is easy to miss from there. `CLAUDE.md` and `README.md` both called it "the tour" — they point at the rules file now. 84702f2

---

eslint.config.ts
vzakharov: No reason beyond it being what `create-next-app` emits. Converted.

Worth doing for more than consistency, as it turns out: `tsconfig.json` includes `**/*.ts`, `**/*.tsx` and `**/*.mts` — so the `.mjs` was the one config in the repo TypeScript never saw. Renaming it surfaced a real error immediately (rule severities widening to `string`, so the config objects didn't satisfy `ConfigWithExtends`); annotating `boundariesConfig` as `Config` from `eslint/config` restores contextual typing and it checks clean. No module shims needed — `eslint-plugin-boundaries` ships its own declarations.

One dependency note: `jiti` is now an explicit devDependency. ESLint needs it to load a TS config but declares it only as an *optional* peer, and it was reaching us transitively through `@tailwindcss/postcss` — a Tailwind bump could have dropped it and quietly broken linting.

Re-verified both FSD guards against the converted config: eslint rejects an upward `features → pages` import and a deep `@/shared/ui/card` import; steiger rejects the upward one. 88aca7b

---

.claude/rules/fsd.md
vzakharov: > Two caveats up front: I can't read Playgramai/playgramapp from this session (cross-owner repo attach is refused here),

wrong premise to start with. pls go ahead and try; you have all the means (you have shimmed gh)

vzakharov: Correcting myself: the "can't read playgramapp" premise was wrong. I inferred it from one `add_repo` error and never tried `gh`, which works fine. I've read it now, and it changes part of my answer.

**What playgramapp actually does.** `src/app` with six segments (`api`, `config`, `providers`, `styles`, `tests`, `ui`), root `app/` as the router — `app/layout.tsx` is 29 lines of CSS-import ordering and metadata, ending in `export { RootLayout as default } from '@/app/providers'`. So your "root `app/` being just thin reexports" is exactly right. And it does run `steiger src/` — with **one** override:

```js
files: ['./src/app/ui/**'],
rules: { 'fsd/no-ui-in-app': 'off' },
```

**So my "two overrides" was wrong** as a description of that pattern's cost. The second one came from my own probe naming a segment `providers`, which I presented as inherent. It isn't — I re-probed one segment at a time: `ui` trips `no-ui-in-app`, `providers` trips `segments-by-purpose`, and `styles`, `config`, `api`, `tests` are all clean.

**But there's a wrinkle**, and it's the reason playgramapp needs no `segments-by-purpose` override despite having `src/app/providers`: it's on `@feature-sliced/steiger-plugin` 0.5.7, and we're on 0.7.0, which added

```js
var BAD_NAMES_REACT = ["hook", "hooks", "context", "provider", "providers"];
```

That string is absent from the 0.5.7 bundle and present in 0.7.0 (I diffed them). So copying playgramapp's layout here verbatim would cost two overrides today — not because it misjudged, but because the plugin got stricter after the fact. Renaming that one segment would put it back to one.

**Was it a misjudgment there? No** — and the real difference isn't the override count, which is a symptom. It's that playgramapp's app layer is a genuine layer: six segments and an entire app shell under `src/app/ui`, with its own tests. One override to place that correctly is a good trade. Ours is three files — root layout, theme provider, `globals.css` — with nothing to segment, so `src/app` would buy indirection and an override and give back nothing.

I've rewritten that section of `.claude/rules/fsd.md` around this rather than the override count, including which segment names actually trip the rule and a pointer to revisit when there's an app shell here worth the same. 1c65a9a

The offer stands: if you'd rather have symmetry with playgramapp than a zero-override config, it's a small change and I'll make it.

---

vzakharov: Reframing accepted — that's a better argument than the one I offered. Symmetry with another repo is worth little; "every layer lives in the same place" is a property of *this* codebase, and five-in-`src/`-one-at-root made it a rule with an exception stapled on. Done. 0fcd7f6

- `app/root-layout.tsx`, `app/theme-provider.tsx`, `app/globals.css` → `src/app/{ui,styles}`
- Root `app/` is routing alone now: `layout.tsx` and every `page.tsx` are one-line re-exports
- `eslint-plugin-boundaries` gains an `app` element — segmented like `shared` rather than sliced, allowed to import every layer through its public API, free to move between its own segments

**Cost: one override**, `fsd/no-ui-in-app` on `src/app/ui/**`. Next mandates a root layout and a root layout is app-layer UI wherever you file it, so the rule genuinely has no answer here — that's the comment in `steiger.config.mjs`. The *second* override stays avoided by segment naming: the theme provider is a file under `src/app/ui/` rather than a `src/app/providers/` segment, since `provider(s)`/`context`/`hook(s)` are on the plugin's list and `styles` isn't.

Both checkers re-verified against an upward `features → app` import; each rejects it. Behaviour unchanged — same routes, same rendered `<head>` on every page, Tailwind still resolving sources from the moved stylesheet. The PR body and squash proposal are updated, including dropping the now-false "no exemptions" claim.

---

**One thing I found while moving it, and did not fix.** `app/layout.tsx` was `export { default } from './root-layout'` — which re-exports only the default. The root layout's `metadata` export has therefore never reached Next: `/` ships with no `<title>`, no description and no Open Graph tags at all.

```
$ grep -oE '<(title|meta)[^>]*>' out/index.html
<meta charSet="utf-8"/>
<meta name="viewport" content="width=device-width, initial-scale=1"/>
<meta name="next-size-adjust" content=""/>
```

It's pre-existing — `main` has the identical `export { default } from './RootLayout'` — so this PR didn't cause it, and I kept the move behaviour-neutral rather than quietly fixing it inside a refactor that claims no behaviour changes. But it sits exactly on the seam I just moved, and it's a one-line fix (`export { RootLayout as default, metadata } from '@/app/ui'`). Want it here, or as its own PR?

---

.claude/skills/sync-agent-boilerplate/source.json
```diff
@@ -28,16 +28,17 @@
...
     "scripts/lib/watch-tick-common.sh",
     "scripts/pr-body.py",
+    "scripts/run-parallel.sh",
     "scripts/vet.sh"
   ],
   "declined": {
-    ".claude/skills/propose-issue/": "issues are opened by hand here, and the backlog is too small to dedupe against — revisit when filing becomes routine",
-    ".claude/skills/audit-github-backlog/": "the backlog is a handful of issues an operator can read in one sitting — revisit when it is large enough that a sweep beats reading it",
+    ".claude/skills/propose-issue/": "issues are opened by hand here, and the backlog is too small to dedupe against \u2014 revisit when filing becomes routine",
+    ".claude/skills/audit-github-backlog/": "the backlog is a handful of issues an operator can read in one sitting \u2014 revisit when it is large enough that a sweep beats reading it",
```
vzakharov: there are some weird `\u2014` shenanigans going on here. If you can't type an em dash, just use a `-`

docs/remove-before-merging/squash-message.md
```diff
@@ -5,41 +5,29 @@ refactor: adopt Feature-Sliced Design, enforced by Steiger (pr #12)
...
+components/, hooks/, lib/, i18n/ — said nothing about what may depend
+on what. Feature-Sliced Design answers that in a form a checker can
+read: code now lives in src/ under shared/, features/, pages/ and
+app/, every slice publishes an index.ts, and imports point downward
+only, through that barrel. Steiger and eslint-plugin-boundaries both
+enforce it, verified non-vacuous against a planted upward import and
+barrel sidestep.
+
+Placement follows from the rules: the theme toggle is a feature
+because both pages use it; single-consumer pieces stay in their page. Every layer sits under src/, the app layer with
```
vzakharov: this line got out of hand

package.json
```diff
@@ -31,6 +31,7 @@
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-boundaries": "^7.2.0",
     "eslint-plugin-prettier": "^5.5.4",
+    "jiti": "^2.7.0",
```
vzakharov: what is this? isn't node able to run ts natively at this point?

.claude/skills/sync-agent-boilerplate/source.json
vzakharov: Fixed in e5b70d4 — rewrote the file so the em dashes are literal UTF-8 characters. They were escapes because the file was last written through a JSON serializer defaulting to ASCII-only output; nothing about JSON required them.

---

docs/remove-before-merging/squash-message.md
vzakharov: Rewrapped in 0f09fc4, along with a refresh recording the base merge. The proposal comment is updated to match.

---

package.json
vzakharov: You are right — removed in e5b70d4.

ESLint 9.39 loads `eslint.config.ts` through Node's own type stripping (unflagged since 22.18 / all of 24), so the explicit dependency bought nothing. Verified rather than assumed: I hid every jiti copy in the tree and ran the lint clean.

Two notes so the removal is not load-bearing in a way it looks: jiti is still resolvable as ESLint's optional peer via `@tailwindcss/node`, and `deploy.yml` pins Node 20 but only ever runs `pnpm build` — it never loads the ESLint config. So the lint works on either path. If you want the floor made explicit instead of incidental, that would mean an `engines.node` entry plus bumping the workflow off Node 20 (EOL since April) — happy to do it, but it changes deploy, so I left it out of this PR.

---

.claude/rules/content.md
```diff
@@ -0,0 +1,59 @@
+---
+description: How long-form markdown under public/content/ becomes a page — the build-time pipeline, the frontmatter contract, and the traps that fail the build
+paths:
+  - public/content/**
+  - src/shared/lib/content/**
```
vzakharov: why do we place it under `shared/lib/content/` and not just `shared/content/` (or maybe `content-delivery`, or a more fitting term, if simply "content" makes it confusing)?

docs/remove-before-merging/squash-message.md
```diff
@@ -15,19 +15,29 @@ enforce it, verified non-vacuous against a planted upward import and
...
-app-layer UI wherever it is filed. Root pages/ must stay, or Next
-finds src/pages and refuses to build.
+because both pages use it; single-consumer pieces stay in their page.
+Every layer sits under src/, the app layer with them, leaving root
+app/ as routing alone — bought with one override, no-ui-in-app, since
+Next mandates a root layout and a root layout is app-layer UI wherever
+it is filed. Root pages/ must stay, or Next finds src/pages and
+refuses to build.
+
+The layers absorb what main landed while this branch was open. The
```
vzakharov: at least the first sentence -- possibly more -- is archaeology

public/content/generated/mermaid/2a9512422a90.dark.svg
vzakharov, line 1: btw -- I didn't ask before, but why are these thus named? are we programmatically processing those ids -- or whatever they are -- somewhere?

.claude/rules/content.md
vzakharov: Fair challenge — and our own rule already answered it against us. `.claude/rules/fsd.md` says a `shared/lib` sub-library graduates to a top-level segment "once it has several consumers and a purpose identity of its own", and the pipeline had both: ten import sites across three layers (`shared/seo`, `app/lib`, `pages/home`, `pages/case-studies`, and the CV route). It was under `lib` because that's where it landed when it moved, not because it belonged there.

Promoted to **`shared/content`** in d4c54ae. That leaves `shared/lib/hydration` — a single hook — as the one real sub-library, which is what the holding area is for, and the rule now says so on both sides rather than only naming the promotion criterion.

On the name: `content` over `content-delivery` because nothing is delivered — there is no server, so the whole pipeline is build-time compilation that resolves into the build graph and is thrown away with it. `content` names the concern the segment serves, the same shape as `seo` and `i18n` (as against `utils` or `markdown`, which would name the contents). Steiger agrees: `pnpm lint:fsd` passes on `shared/content` and rejects a planted `shared/utils` with `fsd/segments-by-purpose`, so that's a real check rather than a vacuous one.

Full vet is green.

---

docs/remove-before-merging/squash-message.md
vzakharov: Agreed — "The layers absorb what main landed while this branch was open" describes the branch, and the squashed commit won't have one. Cut in d4c54ae.

The rest of the paragraph was change-narration too ("becomes", "end up in"), so it's rewritten as present-tense placement with the reason attached:

> The markdown pipeline is shared/content behind one index.ts: build-time-only and site-agnostic, so shared rather than an entity, and a segment of its own rather than a shared/lib sub-library because consumers on three layers reach for it. Its components and both routes are one pages/case-studies slice, since slices on a layer cannot share sideways. …

---

public/content/generated/mermaid/2a9512422a90.dark.svg
vzakharov: Yes — the name is the mechanism, not a UUID. `2a9512422a90` is `sha256(fence.trim())` truncated to 12 hex chars, from `mermaidHash()` in `src/shared/content/mermaid-hash.ts`. Both sides compute it:

- `scripts/render-mermaid.ts` walks every `mermaid` fence under `public/content/`, writes `<hash>.light.svg` / `<hash>.dark.svg`, and prunes renders no fence hashes to any more.
- `rehype-mermaid.ts` hashes the fence again at build time and looks the pair up, throwing if either is missing.

Content-addressing is what buys the staleness guarantee: edit a diagram and its hash moves, so the build fails on the missing render rather than shipping the old picture. A slug- or ordinal-based name would keep pointing at a stale file after an edit, and would need separate bookkeeping to know what to prune.

Confirmed against the two fences in `playgram-bubble-to-nextjs-part-1.md` — they hash to exactly the two names on disk.

The cost is legibility: the directory and its diffs tell you nothing about which document a render belongs to. A `<slug>-<hash>` prefix would fix that without giving up the guarantee (the hash still moves on edit); it just means the same fence text in two documents renders twice. Happy to add it if you want the tree browsable — leaving it as-is for now since it's a judgment call, not a defect.

---

vzakharov:
### Attestation

Verified `8e91458` (`claude/fsd-architecture-adoption-ga5ff0`) against `main` @ `b2924c6` "chore: sync the agent boilerplate forward to eab4dc7 (pr #13)".

- `./scripts/vet.sh` — **green** (`pnpm build`, `typecheck`, `eslint .`, `format:check`, `lint:fsd`, `type-overlap`)
- CI on this PR — **nothing runs on a PR in this repo**; `.github/workflows/deploy.yml` fires only on push to `main`
- CI-only buckets — **not dispatched: this project has no `/test-on-gh` lane**, so there is no bucket beyond what the vet run covers

`/check-merge` reported `contained` against `main` @ `b2924c6` — the base has not moved out from under the branch.

Two caveats a reviewer should weigh, neither of which the vet run can speak to:

- **There is no test suite.** `pnpm build` stands in for one: it type-checks every page, resolves every import and renders every route to static HTML, so it catches breakage — but it says nothing about whether a page is *correct*. The behaviour claim in the squash body ("every route renders as it did, in both themes") rests on the rendered-page comparison done earlier in the branch, not on an assertion anything re-runs.
- **The `metadata` export gap noted in the PR body is still open** — `app/layout.tsx` re-exports only the default, so the root layout's `metadata` never reaches Next and `/` ships without `<title>` or Open Graph tags. Pre-existing on `main`, deliberately preserved rather than silently fixed here.

---

PR #9, opened 2026-08-21
Playgram case study, Part I

vzakharov:
## Summary

- **Turns the brief in #4 into three finished documents** under a new `content/` root: the full case study (~9,100 words of prose), plus genuine rewrites at ~1,900 and ~420 words rather than truncations. No page, no PDF, no route — the issue defers those.
- **Every bracketed placeholder is filled from the Playgram repo's actual history**, not estimated: 1,537 commits, 1,063 merged PRs, 3,487 split files, 362 enabled lint rules (28 hand-written), 8,123 internal imports with zero boundary violations, the real `type-overlap` ratchet sequence, all 33 skills, and a live workflow excerpt from the Bubble split.
- **Six of the brief's claims did not survive checking**, and the articles carry the corrected version of each. Most consequential: the export is documented as **~25 MB**, not 11 MB — that figure is identical across all 87 revisions of the file recording it, and the raw JSON is gitignored so it cannot be measured. Also: the docs-only stretch was four days rather than two weeks; the GCP detour ran Railway → Cloud Run → Railway rather than starting on GCP; the framework was never deliberated; the four-provider model panel is accurate only from 14 March; and Zeroqode's "#1 plugin provider" is unsupportable from any neutral source.
- **Adds a hand-written SVG chart** generated from a committed CSV, so the chart and the prose cannot disagree. It carries the finding the brief asked to have confirmed — the method change is a four-week seam, not a spot, and the number that actually proves the parallelism is that churn per commit stays flat at ~400 lines while commits/day rise 63%.
- **Files three follow-up article issues** (#6, #7, #8) so the deferred depth has a home instead of bloating Part I.

Two deviations worth the reviewer's attention: the full version is ~50% over the planned ~6,000 words, and the tightening pass established that this is substance rather than padding (it recovered 68 words), so getting to 6,000 means cutting researched sections — a scope call left open. And every external citation was replaced after verification; the commonly-cited Bubble case studies (Comet, Dividend Finance) trace only to agency marketing and are not used.

## QA Checklist

- [ ] `render-full` — open `content/case-studies/playgram-bubble-to-nextjs-part-1.md` on GitHub and confirm the Mermaid import-graph diagram, the SVG chart, and the screen-recording embed all render
- [ ] `chart-themes` — view the rendered chart with GitHub set to light and then to dark; the line, annotations and milestone labels must be legible in both, with nothing clipped at the right edge or the baseline
[...]

Issue #4, opened 2026-08-17
Playgram case study Part I (wip)


vzakharov:
Make, first as an in-repo md file (for further conversion into a proper page and PDF once the content is finalized -- not as part of this issue), a case study of my recent work (Mar 9 - Aug 9) converting the Playgram bubble AP to code.

Approximate plan/structure. Notes:
- brackets indicate what needs to be supplemented by the agent themself
- when I use Russian, it's not an indication to use the same in the story -- I was just struggling to find the right idiom, but you won't:)
- although you'll be most certainly extending on my thoughts below, pls try to keep my tone of voice, and certain parts verbatim, even when they sound mildly unnatural (but not if they're way off)
- throughout, pick some quotes you can use as quote callouts (like, something first put in the text is then repeated as a separate quote)
- when we mention specific workspaces, do not include their names, as this repo is public, as well as all its branches/PRs
- the structure/splitting into sections/their order is from my head and does not have to be the same. If you think you can suggest a better structure from the pov of storytelling or otherwise, pls do.
- after writing so much in this brief, I don't even know if all of it belongs in the case study. Maybe some of this is subject to follow-up articles instead. I don't want to lose it because it all seems equally important to me, but of course we need to think from the pov of a reader, and of course the cta (not one we'll mention openly, but one we keep in mind) is "hire this guy"
- as you can see, I ended up making this "Part 1" because of just how much info I'm trying to fit in. I'm just feeling that if I don't, I'll never finish it because writing this brief is taking its toll on me lol

Title: From Bubble to Nextjs in 4 months: Playgram case study

disclaimer: the disclosures in this case study were approved by Playgram management, i.e. no NDA breach

pre-intro: after I started writing this case study, I realized that even the [brief](https://github.com/vzakharov/vovazakharov.com/issues/4) I'll give Claude to hydrate turned out to be X,000 [tbd] words long -- and even then, after I've finished it I understand that I've only scratched the surface of everything I wanted to tell. So, (a) sorry it's so long, (b) check out the mini version, micro version and nano version [we'll have to create all three], and (c) I'll post more detailed insights on some or all of the aspects later.

## intro

in case you didn't know, bubble is a no code app builder

when you think of such tools, you'd think that people would just use it to make simple things like [enumerate a couple]

but you'd be surprised how far people can actually take it -- bubble's been used to create [research and list a few eyebrow-raising use cases; perhaps bubble's own [success stories page](https://bubble.io/customers) is a good place to start]

[...]

PR #13, opened 2026-09-01
chore: sync the agent boilerplate forward to eab4dc7

vzakharov:
## Summary

- **Syncs the agent boilerplate forward from `vzakharov/agent-project-boilerplate`**, from watermark `1f6ee79` to `eab4dc7`. Eight candidate commits triaged; four ported (whole or in part), four skipped. The triage table is below — the watermark advances past the skips, so this PR body is the only record of why.
- **`/issue` no longer owns implementation.** It had restated the whole lifecycle — export, split, implement, open a draft PR — so a session could conclude, correctly per the text, that it needed no plan. It now exports, reads and splits, then hands `/pr` a one-line task plus the issue number to close. A native plan-mode launch is refused, since the flow needs a committed export and plan file that plan mode cannot write. The export gets committed in its own step before any planning, and every slice of a split becomes a sub-issue — the first included, so the PR has a child to close — linked through the sub-issues API rather than by `Part of #<parent>` prose alone.
- **The GitHub exporter reports attachment failures instead of swallowing them.** A failed download was treated as a skip: one stderr line, a success count with no denominator, nothing printed at all when every download failed, and exit 0 — a thread whose screenshots never arrived read exactly like a clean export. It now prints the count as a fraction, names every URL that failed, and exits non-zero, while still writing the Markdown.
- **The sync watermark gains a `{path: note}` form** for paths taken but not verbatim, so the next sync arrives knowing what the local re-expression does differently. Two entries use it: this skill (renamed from the source's `sync-upstream`) and `scripts/vet.sh` (rewritten for this stack).
- **Two paths declined, with conditions a later sync can re-check:** `/bootstrap-workflow-dispatch` (the 404 it clears cannot occur here, and it `@`-references the declined `/test-on-gh`) and `scripts/run-parallel.sh` (`vet.sh` already fans out; the source's own catalog calls the helper optional).
- Incidentally: `scripts/lib/__pycache__/github.cpython-311.pyc` was tracked, so every run of either Python script showed binary churn in `git status`. Untracked, and `__pycache__/` gitignored.

### Triage of the eight candidates since `1f6ee79`

| --- | --- | --- |
| `eab4dc7` stop `/issue` owning implementation | **take** (partial) | Ported; its `/audit-github-backlog` half dropped — that skill is declined here — and, on review, its CLAUDE.md decision-doc criterion too: this repo keeps no decision docs, so `source.json` records CLAUDE.md as a `{path: note}` adoption |
| `d245ef5` report attachment download failures | **take** | This repo had the silent-skip behaviour the commit fixes |
| `efbe3e9` share the GitHub plumbing via `scripts/lib/github.py` | **skip (already have)** + partial take | The lib exists here already; took the `NoReturn` typing on `die` and the import-mechanism docstring |
| `b5a6452` fetch user-attachments past the redirect and proxy | **skip (already have)** | The fix originated in this repo; upstream's message records the patch as deliberately identical |
| `88e23ac` POSIX fan-out check runner for `vet.sh` | **translate** | Took the `{path: note}` watermark form; declined `scripts/run-parallel.sh` itself |
| `5a86cca` adopt `/bootstrap-workflow-dispatch` | **skip (not adopted)** | Declined with its condition recorded — see Summary |
| `585baf3` carry `/plan` and `/implement`'s rationale in the skills | **take** | Stack-agnostic argument about the skills themselves |
| `4370f56` port PR support into the GitHub export script | **skip (already have)** | `export-github-item.py` here already carries PR export, review threads, and both pre-existing bug fixes |

## QA Checklist

- [ ] `exporter-ok` — Export an issue or PR whose thread has an image attachment: `python3 scripts/export-github-item.py <n>`. It writes the Markdown, prints `Downloaded 1/1 attachment(s) under …`, and exits 0.
- [ ] `exporter-fails` — Export a thread with an attachment that cannot download (e.g. run with `GH_TOKEN=invalid`). The Markdown is still written, each failed URL is named on stderr, the count reads as a fraction, and `echo $?` is non-zero. No bearer token appears anywhere in the output.
- [ ] `watermark-noop` — Re-run `/sync-agent-boilerplate`. It clones the source, resolves `eab4dc7`, and reports zero candidates — proving the bump landed on a real SHA and the widened path list still filters.
- [ ] `note-form` — Confirm the sync skill reads a `{path: note}` entry: `.claude/skills/sync-upstream/` and `scripts/vet.sh` are objects in `adopted`, and Step 3's log command still resolves both keys (the run above is the check).
- [ ] `skill-closure` — `bash scripts/check-skill-catalog.sh` passes: every `@.claude/skills/…` reference in the rewritten `/issue`, `/pr` and `/qa-checklist` resolves, and no skill is left a stub.
- [ ] `vet` — `./scripts/vet.sh` is green: `pnpm build`, `typecheck`, `eslint`, `format:check`.
- [ ] `issue-flow` — Read `.claude/skills/issue/SKILL.md` end to end as a fresh session would: the chain line, Step 0's plan-mode refusal, and Step 4's handover agree on where the run ends (a pushed plan file and an `/implement <branch>` block, not code and not a PR).

| --- | --- | --- | --- |
| `exporter-ok` | integration | ❌ | Needs a token and a fixture thread; no test suite exists yet |
| `exporter-fails` | unit | ❌ | `download_attachments` with a stubbed opener is the natural unit test — the first real candidate for a suite |
| `watermark-noop` | — | — | Requires cloning the source; the sync skill is the check |
| `note-form` | unit | ❌ | Parsing `adopted` entries of both shapes is assertable once a suite exists |
| `skill-closure` | script | ✅ | `scripts/check-skill-catalog.sh` — run and green |
| `vet` | script | ✅ | `./scripts/vet.sh` — `/finalize` runs it |
| `issue-flow` | manual-only | — | Judging whether prose reads unambiguously to a fresh agent |

**Coverage gap:** the two exporter rows are the only automatable-and-uncovered items, and `download_attachments` is now the clearest unit-test candidate in the repo — a pure function over a stubbable opener, with a failure path that matters. CLAUDE.md already names `lib/metadata.ts` and the `messages/` catalogs as the first suite candidates; this belongs on that list. No authorization or permission code is touched, so nothing here is a hard gap.

https://claude.ai/code/session_01TPNHkZU7gaBV8vkJsTXJRM


vzakharov:
Proposed squash title/body:

```
chore: sync the agent boilerplate forward to eab4dc7 (pr #13)
```

```
This repo took its agent infrastructure from
vzakharov/agent-project-boilerplate, which keeps editing those files.
Eight commits touched an adopted path since the 1f6ee79 watermark; four
are ported here whole or in part, four are skipped as already-held or
stack-bound. The watermark advances past the skips, so pr #13's body is
the only record of why.

/issue no longer owns implementation. Restating the whole lifecycle let
a session conclude, correctly per the text, that it needed no plan; it
now exports, reads and splits, then hands /pr a one-line task plus the
issue number to close. Every slice of a split becomes a sub-issue, the
first included, linked through the sub-issues API.

The GitHub exporter reports attachment failures instead of swallowing
them — a thread whose screenshots never arrived read exactly like a
clean export. It prints the count as a fraction, names every URL that
failed, and exits non-zero, while still writing the Markdown.

The watermark gains a `{path: note}` entry form for paths taken but not
verbatim, so the next sync expects `translate` rather than `take` on
commits touching them. CLAUDE.md takes that form — the port's
decision-doc guidance stays out, this repo keeping no decision docs —
as do this skill and scripts/vet.sh. /bootstrap-workflow-dispatch and
scripts/run-parallel.sh are declined outright, with conditions a later
sync can re-check.

```

---

docs/remove-before-merging/squash-message.md
vzakharov, line 1: let's cut in ~half

scripts/lib/__pycache__/github.cpython-311.pyc
vzakharov, line 1: let's gitignore this (and all suchlike ofc)

CLAUDE.md
```diff
@@ -140,6 +140,7 @@ The proposed squash title/body goes up when the PR opens and is kept in sync as
...
 - **Decision-doc consistency**: When revising a decision (e.g., renaming a convention, changing a tool choice), update all docs that reference the old convention.
+- **A decision doc records decisions, not implementation.** A line belongs there if it answers _why we chose this_ or _what constraint we accepted_; if it answers _how it works_, it belongs in the code. Don't name a function, module path or column unless the decision is _about_ that name. A named trap ("changing X silently breaks Y") goes in a comment beside X — the person about to break it is reading the code, not the doc. A superseded decision compresses to what was rejected and the one-line why, not a paragraph of mechanics that have since been deleted; a settled fork compresses to the decision it reached, not the comparison that reached it. A decisions index or summary takes the same test: a row is the decision's précis, never its mechanics. The cost isn't only the context every agent spends reading them — a decision survives a refactor and a description of how it works does not, so implementation detail goes stale invisibly and leaves the doc untrustworthy rather than merely long.
```
vzakharov: have we ever had any decision docs here? does this come from the boilerplate? I wouldn't say we need this mumbo-jumbo here

docs/remove-before-merging/squash-message.md
vzakharov: Cut to four paragraphs in 7e1e899 — roughly half the previous length. Gone: the exporter's failure-text mechanics (fraction, body-free reason, S3 echoing the token), the sub-issue API details, the `/pr <issue>` parameter, and the incidental `.pyc` untrack. Each paragraph now names the behaviour that changed and leaves the how to the diff. The `Proposed squash title/body:` comment is PATCHed from the same file.

---

scripts/lib/__pycache__/github.cpython-311.pyc
vzakharov: Already done in this PR — 2e4cf3d untracks the file and adds `__pycache__/` to `.gitignore`, which is what you're seeing in the diff (the file appears because it's being deleted). `__pycache__/` covers all of it: CPython 3 writes every `.pyc` into that directory, never beside the source, so a `*.pyc` line on top would be dead weight. `git ls-files | grep pyc` is empty on the branch head.

---

CLAUDE.md
vzakharov: Yes — it came in with the eab4dc7 port, not from anything here. This repo has no decision docs: `docs/` holds plans and the transient `remove-before-merging/` tree, and the only other hits for the term are prose in a case study. Dropped in 3b0c487.

`source.json` now carries CLAUDE.md as a `{path: note}` adoption recording that this material is deliberately not taken, so the next sync expects `translate` on commits touching it rather than re-offering it as a straight take. The PR body's triage row for eab4dc7 says the same.

---

vzakharov: Follow-up on the sibling bullet above it, which predates this PR (it came in with 196a071, the original adoption). Its rule still applies here — the skills cross-reference each other densely, so a renamed term needs a grep — but its `Decision-doc consistency` label named the same absent doc class. Relabelled to `Convention renames` and reworded to say what it actually asks for, in 706a5e4.

---

vzakharov:
### Attestation

Verified `0b89ab4` (`claude/sync-boilerplate-upstream-cpaz8q`) against `main` @ `e0ccd2b` "feat: #4 write the Playgram case study, Part I (pr #9)".

- `./scripts/vet.sh` — **green** (`pnpm build`, `typecheck`, `eslint`, `format:check`)
- `bash scripts/check-skill-catalog.sh` — **green**: every `@.claude/skills/…` reference in the rewritten `/issue`, `/pr` and `/qa-checklist` resolves, no stubs
- CI on this PR — **nothing runs on a PR in this repo** (`deploy.yml` is push-to-`main` only)
- CI-only buckets — **not dispatched: `/test-on-gh` is declined here** (no test suite, so no CI-only bucket exists — recorded in `source.json`)

The vet run was at `706a5e4`; the two commits since it (`825ff73`, `0b89ab4`) add and then delete `docs/remove-before-merging/squash-message.md`, which no check reads and which is now swept off the branch.

`main` had not moved at the merge check — `e0ccd2b` is still both its head and this branch's merge base, so nothing was merged in.

---

PR #11, opened 2026-09-01
feat: adopt the type-overlap gate from playgramapp

vzakharov:
## Summary

- Ports `Playgramai/playgramapp`'s cross-file duplicate type-member detector as `scripts/type-overlap-check.ts`, **wholesale and at its enforced floor of 1** — any member two named type aliases both declare is a finding, nothing grandfathered, no ratchet to phase in. Detector semantics are unchanged from upstream (pairwise member-set intersection, own members only, property signatures only, `.d.ts` and `*.test.*` skipped, `TYPE_OVERLAP_MIN` upward-only, findings grouped by identical shared member set).
- **Two adaptations, both because this repo's layout differs from upstream's.** Upstream's explicit root allowlist (`src`, `scripts`, `app`, `eslint`) becomes a skip-list over the repo root: the failure modes aren't symmetric — an allowlist silently un-scans the next source directory someone adds, where a skip-list at worst scans one holding no type aliases. And `@typescript-eslint/consistent-type-definitions` now bans `interface`: the gate scans type aliases only, which is complete upstream *because* interfaces are banned there, so without the ban it would be blind to every shape spelled as one. The four existing interfaces convert to type aliases.
- **The repo had exactly one finding at threshold 1**: `ProjectCardProps` and `ArticleCardProps` in `components/Card.tsx` both declared `title: string` and `description: string`. Both now intersect `Summarized` (`Titled & Described`, named in `Card.tsx` where both declarers live) over the bases in the new `lib/typings.ts`, the catalog the gate is satisfied by. A repeated *combination* of bases is itself a duplication the gate can't see — [#17](https://github.com/vzakharov/vovazakharov.com/issues/17) tracks closing that, and a repo-wide sweep found this pair was the only one.
- **`tmp/` is exempt from eslint and prettier**, matching the gate's own skip-list — CLAUDE.md sends every dev artifact there, so a spike must not fail vet on formatting it has no reason to obey.
- Wired into `scripts/vet.sh` alongside typecheck/eslint/format (it reads source text only, so it overlaps them safely). `scripts/type-overlap-check.README.md` — the reference beside the script it documents — carries the rationale, the naming families for a base, the distinctions that resolve differently, and the known blind spots — adapted where upstream leans on machinery this repo lacks (FSD layer rules, knip, the `playgram/*` lint rules).

## QA Checklist

- [ ] `clean-run` — `pnpm type-overlap` on this branch exits 0 and prints `no overlapping type-pairs found (threshold=1)`.
[...]

PR #14, opened 2026-09-01
feat: deliver the Playgram case study through a build-time markdown pipeline

### LINKED ISSUE #10: feat: deliver the playgram case study onto the website
e0ccd2b delivered the case study as an in-repo md; now we have to actually have it available via the website.

The zero-approximation question we need to answer is: how do we deliver content at all? Do we just take the md and re-narrate it as a react component? Or do we have something that "takes" the md and parses it dynamically on the client?

This is a "load-bearing" concern, as it will serve as the basis for all further content to be delivered via the website (hence cold-loaded bundle size could also become an issue).

Ideally we also want to have the original mds downloadable too (store in `public/`?)

All in all, I want both a durable and elegant solution for content delivery on a static website, and the specific playgram case study delivered.

Riding along, a new section in the CV including the Playgram experience.

vzakharov:
## Summary

- **Answers the issue's zero-approximation question — how do we deliver content at all — with: markdown is the source of truth, and `next build` compiles it to HTML once per deploy.** The page components run in Node on the Actions runner, read the `.md` off disk, and render static files. Pages stays a dumb file host.
- **A content page costs zero bytes of client JavaScript beyond the site's existing baseline** — measured on the build output, the article page's bundle is actually smaller than the homepage's, and no parser, highlighter or diagram renderer appears in any chunk. `import 'server-only'` at the top of every `lib/content/` module makes that structural rather than hopeful: importing one from a client component fails the build.
- **The markdown lives in `public/content/`, so the authored file is itself the published artifact** — one copy, served raw at `/content/…`, no generated tree to drift from the source. Relative links resolve identically in the repo, on GitHub, and on the site.
- **Mermaid fences render to committed SVGs keyed by a hash of the fence**, produced by `pnpm content:mermaid` and never by `next build` — so CI never touches puppeteer, and a fence edited without a re-render fails the build instead of shipping a stale diagram.
- **The case study is reachable**: a Case studies block on the homepage and a Case studies section on the CV, plus Playgram as the CV's most recent experience in both locales. The five hand-written experience cards collapse into one `ExperienceCard` driven by an ordered key list, rather than growing a sixth by copy-paste.
- Both mandatory quality passes ran (`/dry` extracted five duplications the plan couldn't foresee; `/tighten-docs` cut 22 pieces of prose), and `/preview` caught four defects invisible from the source — literal backticks around inline code, code blocks that vanished into the page in light mode, diagrams that collapsed the layout until their SVG loaded, and a 17-item table of contents sitting above the title on a narrow viewport.

### Known gap

The 8.3 MB screen recording is **hotlinked** from `github.com/user-attachments` rather than vendored, which is option **2b** in the plan rather than the recommended **2a**. `/user-attachments/` is unreachable from this execution environment — the agent proxy scopes `github.com` to repository endpoints and refuses it — so there was nothing to download and transcode. It renders as a player either way; the cost is the plan's stated one (the asset serves as `video/quicktime`, which Firefox will not play) and vendoring later is a one-line edit to the markdown.

## QA Checklist

- [ ] `index` — `/case-studies` lists the piece with its date, reading time and part marker, and links to the mini and micro cuts
- [ ] `article` — the full article renders end to end: title, description, outline, prose
- [ ] `variants` — the Full/Mini/Micro switcher moves between the three cuts and marks the current one
- [ ] `cross-links` — the "mini version"/"micro version" links *inside the prose* go to site routes, not to raw `.md` files
- [ ] `raw-md` — the "Markdown" link serves the authored file at `/content/case-studies/…​.md`, frontmatter and all
- [ ] `diagrams` — both Mermaid diagrams render, are legible, and swap when the theme toggle flips
- [ ] `code` — code blocks are syntax-highlighted and sit on a visible surface in **both** themes
- [ ] `tables` — the wide comparison tables scroll inside their own container; the page never scrolls sideways
- [ ] `video` — the screen recording renders as a player and plays (needs real network; the asset is hotlinked)
- [ ] `narrow` — at phone and tablet width the article title comes before the table of contents, not after it
- [ ] `home` — the homepage `/dev` section shows a Case studies block above Featured Projects, and everything else there is unchanged
- [ ] `cv` — `/en/cv` and `/ru/cv` both show Playgram first under Experience and a Case studies section linking to the article; every pre-existing entry is intact
- [ ] `cv-print` — the CV's ⎙ / PDF view is unchanged apart from the new entries
- [ ] `mermaid-gate` — edit a `mermaid` fence without re-rendering and confirm `pnpm build` fails naming the fence
- [ ] `sitemap` — `/sitemap.xml` lists `/case-studies` and all three document routes

| `index` | integration | ❌ | Render the index page, assert one card per full document with its derived title and reading time |
| `article` | e2e | ❌ | Build, then assert the article HTML carries the title, the outline and the body |
| `variants` | integration | ❌ | `generateStaticParams` yields base + each variant; each route renders its own file |
| `cross-links` | unit | ❌ | `rehypeContentLinks` over a fixture: `./x.mini.md` → `/case-studies/x/mini` |
| `raw-md` | e2e | ❌ | Assert `out/content/case-studies/*.md` exists and the header's link points at it |
| `diagrams` | unit | ❌ | `rehypeMermaid` emits both renders and throws on a hash with no SVG — the throw is the load-bearing half |
| `code` | manual-only | — | Whether a surface reads as distinct from the page is a judgement call; this is the defect `/preview` caught |
| `tables` | unit | ❌ | `rehypeTableScroll` wraps each `<table>` exactly once and does not re-wrap its own wrapper |
| `video` | manual-only | — | Live third-party asset over the public internet; not reachable from CI or from this environment |
| `narrow` | manual-only | — | Reflow at a real mobile viewport; headless Chrome clamps below 500px, so it needs a human or Playwright |
| `home` | integration | ❌ | Snapshot the homepage's text content against `main` — the check that caught nothing was lost in the de-`prose` pass |
| `cv` | integration | ❌ | Same snapshot for `/en/cv` and `/ru/cv`; this is what verified the `ExperienceCard` refactor lost nothing |
| `cv-print` | manual-only | — | Print stylesheet output needs a human looking at a PDF |
| `mermaid-gate` | integration | ❌ | Point the plugin at a fixture with no matching render and assert it throws |
| `sitemap` | unit | ❌ | `sitemap()` returns the static routes plus one entry per document |

Every automatable row is uncovered because **the project has no test suite yet** (CLAUDE.md § Testing — `pnpm build` stands in for one). This change adds the first logic genuinely worth asserting on, so `lib/content/` is a strong candidate for the suite's first home; that's follow-up work, not this PR.

Closes #10

https://claude.ai/code/session_01R8WuuV5chLRQAUQhwAUwku


vzakharov:
Proposed squash title/body:

```
feat: #10 publish case studies from build-time-compiled markdown (pr #14)
```

```
The site had a finished case study in the repo and no way to put it in
front of a reader. This settles how it delivers prose, for every document
that follows: markdown under `public/content/` is the source of truth and
`next build` compiles it to HTML, so the authored file is itself the
published artifact — served raw at `/content/`, its relative links
resolving the same in the repo, on GitHub and on the site.

A content page therefore costs zero bytes of client JavaScript beyond the
site's baseline, which `import 'server-only'` atop every `lib/content/`
module holds by construction. A new document is a page and a sitemap entry
with no route work, the collection registry being the one place a content
URL is shaped. Mermaid fences are the single authoring step:
`pnpm content:mermaid` renders each to a committed pair of SVGs keyed by a
hash of the fence, and the build fails on one whose render is missing
rather than shipping a stale diagram.

Riding along, the CV gains Playgram in both locales, and its five
hand-written experience cards collapse into one component. The pipeline's
types compose the repo's shared bases, `WithOptionalClassName` joining
`lib/typings.ts` for the one member that spans the site rather than the
pipeline. The screen recording is hotlinked from GitHub's
user-attachments rather than vendored, so it serves as `video/quicktime`
and Firefox will not play it.

Closes #10

```

---

components/content/ArticleBody.tsx
```diff
@@ -0,0 +1,13 @@
+export interface ArticleBodyProps {
+  /** HTML from `renderDocument` — first-party markdown, compiled at build time. */
+  html: string;
+}
+
+export function ArticleBody({ html }: ArticleBodyProps) {
+  return (
+    <div
+      className="prose prose-content max-w-none"
+      dangerouslySetInnerHTML={{ __html: html }}
```
vzakharov, line 7: are there any alternatives that don't involve setting it this way? don't edit yet, pls advise

lib/content/frontmatter.ts
```diff
@@ -0,0 +1,37 @@
...
+  const result = frontmatterSchema.safeParse(data);
+
+  if (!result.success) {
+    throw new Error(
+      `Invalid frontmatter in ${fileName}:\n${z.prettifyError(result.error)}`
+    );
+  }
+
+  return result.data;
+}
```
vzakharov: why not just .parse (and maybe not introducing a wrapper at all)?

lib/content/hast-text.ts
vzakharov, line 1: what's "hast" supposed to mean? Is it some standard term/jargon?

lib/content/mermaid-hash.ts
vzakharov, line 1: why not .ts

messages/ru.json
```diff
@@ -31,7 +31,36 @@
     },
     "experience": {
       "title": "Опыт",
-      "project1": {
+      "playgram": {
+        "title": "Разработчик",
+        "period": "Playgram (март – август 2026)",
+        "description": "Перенёс работающий AI-чат с богатой функциональностью — несколько провайдеров и моделей, командные чаты в реальном времени, библиотеки изображений и файлов, память и управление знаниями, голосовой ввод — с no-code-платформы Bubble в продакшн-кодовую базу на Next.js 16, не выводя продукт из эксплуатации.",
```
vzakharov: "AI-чат" не звучит как класс платформы

messages/ru.json
```diff
@@ -31,7 +31,36 @@
...
+        "intro": "Сделано за 158 дней: двадцать с лишним AI-агентов работали параллельно, а я ревьюил их работу, а не печатал код:",
```
vzakharov: "а не печатал код" is AI-speak; just say ": паралелльные AI-сессии до 20 одновременно работающих агентов"

messages/ru.json
```diff
@@ -31,7 +31,36 @@
...
+        "items": [
+          {
+            "label": "1 029 смерженных пул-реквестов",
+            "text": "и 250 000 строк продакшн-TypeScript, почти ничего из этого не написано руками."
```
vzakharov: ", почти" -- не нужно

messages/ru.json
```diff
@@ -31,7 +31,36 @@
...
       "title": "Опыт",
-      "project1": {
+      "playgram": {
+        "title": "Разработчик",
+        "period": "Playgram (март – август 2026)",
+        "description": "Перенёс работающий AI-чат с богатой функциональностью — несколько провайдеров и моделей, командные чаты в реальном времени, библиотеки изображений и файлов, память и управление знаниями, голосовой ввод — с no-code-платформы Bubble в продакшн-кодовую базу на Next.js 16, не выводя продукт из эксплуатации.",
+        "intro": "Сделано за 158 дней: двадцать с лишним AI-агентов работали параллельно, а я ревьюил их работу, а не печатал код:",
+        "items": [
+          {
+            "label": "1 029 смерженных пул-реквестов",
```
vzakharov: name the total number of main commits (this is more load-bearing, since in pre-PR commits to main were still gestalts)

messages/ru.json
```diff
@@ -31,7 +31,36 @@
...
+            "text": "и 250 000 строк продакшн-TypeScript, почти ничего из этого не написано руками."
+          },
+          {
+            "label": "48 релизов и 18 хотфиксов:",
+            "text": "деплой в продакшн каждые 2,4 дня, три переезда рабочих пространств, ни одного отката."
```
vzakharov: ", три переезда" -- unneeded

messages/ru.json
```diff
@@ -31,7 +31,36 @@
...
+          },
+          {
+            "label": "Холодная загрузка меньше секунды",
+            "text": "вместо нескольких секунд."
+          },
+          {
+            "label": "Feature-sliced-архитектура, закреплённая линтерами,",
+            "text": "чтобы толпа агентов не расшатала структуру: из 8 123 импортов ни один не идёт вверх и ни один — вбок."
```
vzakharov: ditch the part after colon

messages/ru.json
```diff
@@ -120,6 +149,14 @@
         ]
       }
     },
+    "caseStudies": {
+      "title": "Разборы проектов",
+      "playgram": {
+        "title": "From Bubble to Next.js in 4 months: the Playgram case study",
+        "description": "Часть I из II — подробный разбор переезда, описанного выше: почему стоило уходить с no-code, какая архитектура и какие инструменты позволили двадцати агентам работать одновременно, и все цифры. На английском.",
```
vzakharov: "Пока на английском"

public/content/case-studies/playgram-bubble-to-nextjs-part-1.md
```diff
@@ -285,6 +291,12 @@ As an example, here's the real import graph of the final state — every arrow i
 
 ```mermaid
 flowchart TD
+    accDescr {
+      The final codebase's import graph, six layers deep and acyclic. app sits at
+      the top, then pages, widgets, features, entities, and shared at the bottom.
+      Every arrow points downward and is labelled with the number of import
+      statements crossing it; shared receives the most by a wide margin.
```
vzakharov, line 298: this goes to svg's alt?

public/content/generated/mermaid/2a9512422a90.dark.svg
vzakharov, line 1: let's put a link to the original md in the alt -- in case any agents read the html and can't view images

scripts/render-mermaid.ts
vzakharov, line 1: .ts?

CLAUDE.md
```diff
@@ -32,6 +32,54 @@ Merging to `main` triggers `.github/workflows/deploy.yml`, which builds the stat
 
 Nothing runs on pull requests. `./scripts/vet.sh` runs the same `pnpm build` the deploy does, so a green vet locally is the only pre-merge signal there is.
 
+## Content
```
vzakharov: I feel this is too much for CLAUDE.md; see if you can use `.claude/rules`, where you can bind these to specific parts of the codebase

package.json
```diff
@@ -9,21 +9,40 @@
...
+    "rehype-slug": "^6.0.0",
+    "rehype-stringify": "^10.0.1",
+    "remark-gfm": "^4.0.1",
+    "remark-parse": "^11.0.0",
+    "remark-rehype": "^11.1.2",
+    "server-only": "^0.0.1",
+    "shiki": "^4.4.3",
+    "unified": "^11.0.5",
+    "unist-util-visit": "^5.1.0",
+    "zod": "^4.5.4"
```
vzakharov, line 37: question -- not arguing -- why are these not _dev_ dependencies? don't we only use them during build?

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+```
+The site had a finished case study in the repo and no way to put it in
+front of a reader. The question underneath that — how this site delivers
+prose at all — is answered here once, for every document that follows:
+markdown is the source of truth, and `next build` compiles it to HTML.
+Under `output: 'export'` the page components run in Node on the Actions
+runner, read the `.md` off disk, and render static files; nothing about
+this needs a server, and Pages stays a dumb file host.
```
vzakharov: this sounds like arguing vs my mid-session comment, not as a durable contract

docs/remove-before-merging/squash-message.md
vzakharov, line 1: cut in ~half

components/content/ArticleBody.tsx
vzakharov: Not edited — advising, as asked.

There is exactly one real alternative: **`rehype-react`** in place of `rehype-stringify`. The pipeline stops at hast and compiles it to React elements, so `renderDocument` returns a `ReactNode` and `ArticleBody` renders `{content}` with no `dangerouslySetInnerHTML` anywhere. It stays a server component, so the zero-JS property survives.

What it actually buys, beyond deleting the scary-named call:

- A component map. `a` → `next/link` for internal routes, `pre` → a real `<CodeBlock>` with a copy button, `table` → the scroll wrapper. Three of the five rehype plugins we have are doing tree surgery that a component map would express declaratively.

What it costs:

- Payload. Next emits an RSC flight file per route even under `output: 'export'`, and an 8,000-word article serializes far larger as an element tree than as one HTML string. That's the real bill, and it lands on the page we most want light.
- A dependency and a component map to keep in step with the markdown people actually write.

What it does **not** buy is safety, which is the thing the API name implies. The input is markdown files in this repo, compiled at build time; there is no untrusted content in the path and no user-submitted anything on this site. If that ever changes, the fix is `rehype-sanitize`, not a different render call — swapping to `rehype-react` without sanitizing would keep every injection route open, because `rehype-raw` has already turned the raw HTML into real elements by then.

The two non-options, for completeness: MDX makes documents JSX-capable but stops them being plain markdown GitHub renders, which is the property `public/content/` exists for; `html-react-parser` parses in the browser and costs the zero-JS baseline outright.

**My recommendation: leave it.** Today it would be a payload increase in exchange for an API rename. The moment we want a component for something — copy buttons, `next/link`, a figure with a caption — `rehype-react` pays for itself, and that's the point to switch, with the plugins collapsing into the component map in the same move.

---

lib/content/frontmatter.ts
vzakharov: Done in 42adf20 — the wrapper is gone and `readDocument` calls `frontmatterSchema.parse` directly.

The only thing it was buying was naming the offending file, which a `ZodError` can't know. That's now a rethrow around the read, which covers the YAML parse too — previously a malformed frontmatter *block* (as against a valid block with wrong fields) threw without naming the file either.

---

lib/content/hast-text.ts
vzakharov: Standard jargon, yes — "hast" is the unified ecosystem's HTML syntax tree, the shape every `rehype` plugin operates on. Its siblings are "mdast" (markdown, what `remark` sees) and "unist", the base spec both extend. It's also literally the package the types come from: `import type { Element, Root } from 'hast'`.

Added a line saying so where the reader meets it (42adf20), rather than renaming — the name is the ecosystem's, and renaming would hide the connection to the `hast`/`@types/hast` import right above it.

---

lib/content/mermaid-hash.ts
vzakharov: No good reason any more — both are `.ts` as of 42adf20.

They were JavaScript because `scripts/render-mermaid.ts` runs under bare `node`, never through the bundler, and this module is shared between that script and the build. Node has stripped types on its own since 22.18, so that reason expired. Two details fell out of the move:

- Both importers now spell the extension (`../mermaid-hash.ts`), because Node's ESM resolver won't guess it — hence `allowImportingTsExtensions` in `tsconfig.json`.
- `"type": "module"` in `package.json`, otherwise Node warns on every run that it had to reparse the file as ESM. No `.js` file exists in the repo for the flag to reinterpret, so it's inert beyond that.

---

messages/ru.json
vzakharov: Изменено в 75c2587: «работающую многофункциональную платформу AI-чатов». Заодно Bubble стал «no-code-конструктором», чтобы «платформа» не шла дважды в одном предложении.

---

messages/ru.json
vzakharov: Изменено в 75c2587: «Сделано за 158 дней — параллельные AI-сессии, до 20 одновременно работающих агентов:». Тире вместо первого двоеточия, иначе их два подряд.

То же в английской версии — там было ровно то же самое «rather than typing it».

---

messages/ru.json
vzakharov: Убрано в 75c2587 — «ничего из этого не написано руками». Английское «almost none of it» убрано тем же коммитом.

---

messages/ru.json
vzakharov: Добавлено в 75c2587: **«1 395 коммитов в main,»** / «из них 1 029 — смерженные пул-реквесты; 250 000 строк продакшн-TypeScript, ничего из этого не написано руками.»

Число из самого разбора: 1 395 единиц работы в `main`, 1 029 из них пришли пул-реквестами, остальные — прямые коммиты доPR-эпохи. Так что «1 029 пул-реквестов» действительно недосчитывало то, что было сделано. Английская версия — «1,395 commits on main, 1,029 of them merged pull requests».

---

messages/ru.json
vzakharov: Убрано в 75c2587 — «деплой в продакшн каждые 2,4 дня, ни одного отката.» И «across three workspace cutovers» из английской версии.

---

messages/ru.json
vzakharov: Убрано в 75c2587 — «чтобы толпа агентов не расшатала структуру.» Цифры про 8 123 импорта остаются в самом разборе, где им и место. То же сокращение в английской версии.

---

messages/ru.json
vzakharov: Изменено в 75c2587 — «…и все цифры. Пока на английском.»

---

public/content/case-studies/playgram-bubble-to-nextjs-part-1.md
vzakharov: Yes. Mermaid renders `accDescr` into the SVG's own `<desc>`, but the page embeds the diagram as ``, and an `` hides the SVG's internals from assistive technology — so `rehype-mermaid` lifts the `accDescr` text out of the fence and onto the ``. A fence without one fails the build, which is why every diagram has the block.

As of 42adf20 the alt also carries the source markdown's URL, per your comment on the SVG.

---

public/content/generated/mermaid/2a9512422a90.dark.svg
vzakharov: Сделано в 42adf20, только не внутри SVG, а в `alt` картинки на странице:

```
alt="The final codebase's import graph, six layers deep and acyclic. … Mermaid source: https://vovazakharov.com/content/case-studies/playgram-bubble-to-nextjs-part-1.md"
```

`rehype-mermaid` берёт URL из документа, который рендерится, и приклеивает к описанию из `accDescr`. Почему не в сам файл: SVG именуется хешем фенса, а не документом — один и тот же файл обслужил бы любой документ с той же диаграммой, так что «исходник» у него не один. И читает агент всё равно HTML, а `alt` — это то место, куда он смотрит вместо картинки.

---

scripts/render-mermaid.ts
vzakharov: Yes — `scripts/render-mermaid.ts` as of 42adf20, running on Node's own type stripping (22.18+, noted in the file header). See the reply on `mermaid-hash` for the two mechanical details that came with it.

One change to the code itself: `fs.globSync` became a small readdir scan, since the script is typechecked now and `globSync` is newer than the `@types/node` we're pinned to.

---

CLAUDE.md
vzakharov: Согласен — перенесено в bc6eb85: `.claude/rules/content.md`, привязано к `public/content/**`, `lib/content/**`, `components/content/**`, `app/case-studies/**` и `scripts/render-mermaid.ts`. Правило подгружается, только когда сессия трогает эти пути.

В CLAUDE.md от раздела осталась одна строка под таблицей layout — про то, что такие конвенции живут в `.claude/rules/`. README самой директории поправлен: он утверждал, что она специально пустая, что теперь неправда.

---

package.json
vzakharov: Fair question, and the premise is right: nothing here runs at runtime. Under `output: 'export'` there is no runtime at all, so `next` and `react` are build-only in exactly the same sense — the split can't mean "runtime vs build" in this repo, because nothing would be left on the runtime side.

What the split does mean here is "imported by application source" vs "only reached through config or the toolchain". `shiki`, `unified`, the remark/rehype plugins, `gray-matter` and `zod` are imported from `lib/content/`, so they sit with `next`, `react` and `next-intl`; `tailwindcss`, `@tailwindcss/postcss`, `eslint`, `prettier`, `typescript` and the `@types/*` are only named in config files, so they're dev.

It makes no practical difference either way — CI installs everything (`pnpm install --frozen-lockfile`), and a `--prod` install couldn't build this site regardless, since `tailwindcss` is a devDependency. So this is convention, not mechanism. Happy to move them if you'd prefer the split to read as "build-only"; the consistent version of that move takes `next` and `react` with it, which is why I left it alone.

---

docs/remove-before-merging/squash-message.md
vzakharov: Согласен — переписано в bc6eb85. Было «nothing about this needs a server, and Pages stays a dumb file host», стало описание того, что сделано: markdown — источник правды, `next build` компилирует его в HTML, авторский файл сам является публикуемым артефактом.

---

docs/remove-before-merging/squash-message.md
vzakharov: Сокращено примерно вдвое в bc6eb85 — 36 строк прозы стали 19, четыре абзаца тремя.

---

components/content/ArticleBody.tsx
vzakharov: > Payload. Next emits an RSC flight file per route even under output: 'export', and an 8,000-word article serializes far larger as an element tree than as one HTML string. That's the real bill, and it lands on the page we most want light.

when does it happen? at build time only, or?

messages/ru.json
vzakharov: > из них 1 029 — смерженные пул-реквесты

this is unneeded info at all.

messages/ru.json
vzakharov: no rollback is unneeded either

components/content/ArticleBody.tsx
vzakharov: Written at build time, fetched at runtime — but only on some visits, and I overstated the cost above. Numbers off the current build, for the full article:

| | bytes |
|---|---|
| `playgram-bubble-to-nextjs-part-1.html` | 228 KB |
| `playgram-bubble-to-nextjs-part-1.txt` (the flight payload) | 108 KB |

The `.txt` is the RSC payload, written into `out/` next to the HTML. **A visitor who lands on the URL directly never fetches it** — they get the HTML and that's the whole page. It's fetched only on a client-side navigation into the article (a `<Link>` from the index or the CV), and on prefetch when such a link scrolls into view. So it's a second copy of the page, paid by people who arrive from elsewhere on the site.

Today that payload carries the body as one escaped HTML string — the `dangerouslySetInnerHTML` prop is literally in there. Under `rehype-react` it would instead carry the element tree: ~1,200 elements, each a `["$","tag",null,{…}]` tuple with quoted, escaped prop names. I haven't measured it (that needs the swap actually done), but the scaffolding is on the order of tens of KB on a 108 KB payload — call it a fraction more, not a doubling, and nothing on the direct-load path.

So the cost is real but smaller than I made it sound. It doesn't change the recommendation — there's still no component we want today — but "the real bill" was the wrong framing for a soft-navigation-only delta. If you'd like, I can do the swap on a scratch branch and post the actual before/after sizes.

---

messages/ru.json
vzakharov: Убрано в bb0b3ca — осталось **«1 395 коммитов в main»** / «и 250 000 строк продакшн-TypeScript, ничего из этого не написано руками.» То же в английской версии.

---

messages/ru.json
vzakharov: Тоже убрано в bb0b3ca — «деплой в продакшн каждые 2,4 дня.» И «with no rollbacks» из английской версии.

---

components/content/ArticleBody.tsx
vzakharov: > If you'd like, I can do the swap on a scratch branch and post the actual before/after sizes.

yes pls

vzakharov: Done — swapped `rehype-stringify` for `rehype-react` on a scratch branch off bb0b3ca, built both, then threw the branch away. Numbers are for the full article.

**First, a correction to what I said above.** I claimed a direct landing never fetches the payload. That's wrong: Next inlines the flight payload into the HTML itself, as `self.__next_f.push(…)` script chunks, so hydration has it without a second request. The standalone `.txt` is the *additional* copy used for client-side navigation. So the payload is paid on every article load, not only on soft ones — which makes it a bigger deal than I said, and the measured cost smaller than I implied. Splitting the HTML by that boundary:

| | before | after | delta |
|---|---:|---:|---:|
| **article HTML, total** | 234,151 | 264,075 | **+29,924 (+13%)** |
| — visible markup | 96,621 | 98,548 | +1,927 (+2%) |
| — inlined flight payload | 136,730 | 164,727 | +27,997 (+20%) |
| **article HTML, gzipped** | 65,954 | 70,424 | **+4,470 (+7%)** |
| standalone `.txt` (soft nav) | 111,145 | 145,236 | +34,091 (+31%) |
| mini `.txt` | 36,111 | 39,727 | +3,616 (+10%) |
| micro `.txt` | 23,456 | 24,401 | +945 (+4%) |

So: **+4.5 KB gzipped on the wire for anyone opening the article**, and about the same again for anyone who navigates to it from the index. My "+tens of KB, a fraction more" guess was right on the raw bytes and, once compressed, the whole thing is smaller than one of the page's own SVGs.

**The page comes out equivalent.** Same 6 mermaid `figure`/`img` nodes with their source-URL alts, same 4 table-scroll wrappers, same 376 shiki spans, the video player intact. The only visible-markup change is React escaping more aggressively — 225 extra `&#x27;` and 90 `&quot;`, which is the entire +1.9 KB.

**One side effect worth knowing before we'd ever do this for real.** `rehype-react` augments unified's `CompileResultMap` with `ReactNode`, and `ReactNode` is wide enough to swallow a hast `Root`. So `Plugin<[], Root>` stops resolving as a transformer and starts resolving as a *compiler*, and all five of our plugins lose their tree typing at once:

```
lib/content/plugins/rehype-content-links.ts(80,11): error TS7006: Parameter 'tree' implicitly has an 'any' type.
… ×5
```

The fix is a one-word annotation per plugin (`return (tree: Root) => …`), which is what I did to get the build green. Cheap, but note it's a repo-wide consequence of installing the package, not something contained to the file that imports it.

**Verdict unchanged, with better grounds.** 4.5 KB gzip isn't a reason to refuse; it's also not free, and right now we'd be spending it purely to rename an API — there's no component map, copy button or `next/link` on the other side of it yet. When we want one of those, the trade flips, and three of the five plugins collapse into the component map in the same move. Happy to open it as a follow-up issue so the measurement isn't lost.

---

vzakharov: Filed as [#15](https://github.com/vzakharov/vovazakharov.com/issues/15) — the measurement table, the `CompileResultMap` trap, the trigger condition ("when we want the first component, not before"), and the two non-options are all recorded there, so nothing here is lost when this thread closes.

No labels: the repo's other issues carry none, and there's no tech-debt label to reach for. Say the word if you'd like one created.

---

vzakharov:
### Attestation

Verified `97dfe33` (`claude/playgram-case-study-website-1ztqf8`) against `main` @ `ae7efde` "chore: adopt the type-overlap gate from playgramapp (pr #11)" — merged in and contained, as of this check.

- `./scripts/vet.sh` — **green**, all five checks: `next build` (static export, 12 routes), `tsc --noEmit`, `eslint .`, `prettier --check`, and `pnpm type-overlap`.
- CI on this PR — **nothing runs on a PR in this repo** (`.github/workflows/deploy.yml` triggers only on `push` to `main` and `workflow_dispatch`).
- CI-only buckets — **not dispatched: this project has no `/test-on-gh` skill at all**, not even a stub, so there is no credentialed or browser bucket to reach.

**Re-verified after the base advanced.** `main` moved from `e0ccd2b` to `ae7efde`, which adopted the type-overlap gate — an advance that interacted with this branch three ways rather than sitting beside it, so the merge was not a formality:

1. The incoming `@typescript-eslint/consistent-type-definitions` rule bans `interface` repo-wide, and this branch's new pipeline declared 14 of them. Converted to type aliases.
2. That conversion is what brings those shapes into `pnpm type-overlap`'s scope — it scans type aliases only — which surfaced **11 overlap groups**. Cleared by intersecting a shared base homed at the module owning the member (`DocumentRef`, `WithFrontmatter`, `WithContentDocument`, `WithHtml`/`WithHeadings`/`WithReadingMinutes`), reusing the incoming `Titled`/`WithId`, and adding `WithOptionalClassName` to `lib/typings.ts` as the one member that spans the site rather than the pipeline. Three prop aliases that collapsed to a single base were deleted.
3. One conflict, in CLAUDE.md's repository-layout table, which both sides extended; resolved as a union of both sides' rows.

Note that a green vet here means every route builds, type-checks and is free of duplicated type members — it does not assert the case-study pages are visually *correct*, since this project has no test suite (`pnpm build` stands in for one). The dedup is type-level only and erases at compile time, so it cannot change rendered output.

Working artifacts swept off the branch: `docs/issue/10/`, `docs/plans/10.completed.md`, and `docs/remove-before-merging/` (restored once to reconcile the squash proposal against the merged base, then swept again). `main` tracks no `docs/` files and the branch's net diff under `docs/` is empty, so the add-then-delete pairs cancel out in the squash.

---

PR #16, opened 2026-09-01
feat: adopt a strict linting ruleset

vzakharov:
## Summary

- **Replaces the 30-line `eslint.config.mjs` with an explicit ruleset** — every rule listed as `error` or `off`, never `warn`, each `off` carrying its rationale. `eslint.config.ts` (loaded via jiti) orchestrates the preset spreads and plugin registration; the rules live in `eslint/rule-groups/*.ts`, split by plugin family. Adds `@eslint-react`, `unicorn`, `react-compiler`, all 39 `jsx-a11y` rules and `simple-import-sort`, plus type-aware linting via `parserOptions.projectService`.
- **Nine project-local rules land under a `vova/` namespace.** Rules premised on a database, server actions, Mantine or Zod stay out — a static export gives them nothing to bite on. `no-uncaused-rethrow` keeps its check and drops the suggestion that injected a logging helper this project doesn't have; `no-hardcoded-strings` points at `messages/*.json` and next-intl instead of a per-slice `texts.ts`.
- **The violations it surfaced were mostly real defects.** The theme toggle rendered an English `aria-label` on the Russian CV; `{stars && …}` rendered a bare `0`; three Spotify embeds had no `title`. The locale picker had the same English-label defect one step out of the rule's reach (a template expression, not a literal), fixed alongside. The catalogs are now keyed and typed, and registered with next-intl in `src/shared/i18n/next-intl.d.ts`, so `useTranslations` keys are checked, `useMessages()` returns the real shape, and a key added to `en.json` but not `ru.json` is a type error — which also retired 18 `no-unsafe-*` reports at their source by replacing `t.raw()`.
- **`tsconfig.json` gains the 12 strictness flags several rule-group `off`s delegate to**, so those exemptions are real coverage rather than silent gaps. `eslint-plugin-prettier` is dropped — formatting was checked twice and `pnpm format:check` was always the check doing the work — and `trailingComma` becomes `"all"`.
- **Exemptions sit at the point of use**, per the severity policy now recorded in `.claude/rules/eslint.md`: an inline disable carrying its rationale, where whoever edits the line will read it, rather than softening a rule to `warn`. A config block is reserved for an exemption spanning files, which leaves one — for `eslint/` itself.

## QA Checklist

- [ ] `cv-en` — Open `/en/cv`. The whole CV renders in English, including every bulleted list under Experience and Tech Stack (these moved from `t.raw()` to the typed catalog, so an empty list would mean the rewrite dropped them).
- [ ] `cv-ru` — Open `/ru/cv`. The same page renders in Russian, lists included, and `/cv` still redirects into `/en/cv`.
- [ ] `theme-label` — On `/ru/cv`, inspect the theme-toggle button: its `aria-label` reads "Переключить тему", not "Toggle theme". On `/en/cv` it reads "Toggle theme".
- [ ] `locale-label` — On `/ru/cv`, inspect the flag button: its `aria-label` reads "Переключить на английский". On `/en/cv` it reads "Switch to Russian". Clicking it still swaps locale and keeps you on the CV.
- [ ] `theme-cycle` — Click the theme toggle repeatedly on either page. It cycles light → dark → system → light, and the icon (sun / moon / monitor) matches the active theme at each step.
- [ ] `home-cards` — Open `/`. Project cards that have a star count still show it; cards without one show no star row, and every card's tech-stack line still renders.
- [ ] `home-embeds` — On `/`, the three Spotify embeds under Active Projects still load and play.
- [ ] `og-tags` — View source on `/` and `/en/cv`. The `og:image` is `/ava.png` on the home page with `og:image:width`/`height` of 150, and `/cv_card.png` on the CV **without** width/height; `og:title` and `twitter:title` are unchanged from `main`.
- [ ] `lint-bite` — Add a hardcoded `title="test"` to any JSX in `src/pages/cv/ui/cv-page.tsx` and run `pnpm exec eslint .`; it fails with `vova/no-hardcoded-strings`. Add the same to `src/pages/home/ui/home-page.tsx` and it passes (the scoped `off`). Revert both.
- [ ] `catalog-drift` — Delete any key from `src/shared/i18n/messages/ru.json` and run `pnpm typecheck`; it fails. Revert.

| `cv-en` | e2e | ❌ | Render `/en/cv` from the export, assert a known list item from `cv.techStack.backend.items` is present |
| `cv-ru` | e2e | ❌ | Same against `/ru/cv`, asserting the Cyrillic counterpart; also assert `/cv` redirects |
| `theme-label` | e2e | ❌ | Needs a client render — the label is absent from the SSR markup until `useMounted` flips |
| `locale-label` | e2e | ❌ | Same client-render constraint; assert the label flips with the active locale |
| `theme-cycle` | unit | ❌ | Assert `NEXT_THEME` cycles all three and `ICON` maps each to the right glyph |
| `home-cards` | unit | ❌ | `ProjectCard` with `stars={0}` renders the star row; without `stars`, it does not |
| `home-embeds` | manual-only | — | Live third-party embed; only a human can confirm it loads and plays |
| `og-tags` | integration | ❌ | Assert `constructMetadata()` output for the with/without-`ogImage` cases |
| `lint-bite` | integration | ❌ | A `RuleTester` suite over `vova/no-hardcoded-strings` |
| `catalog-drift` | unit | ❌ | A `tsd`-style expect-error over a catalog missing a key |

The suite covers `scripts/type-overlap-check.ts` and nothing else (CLAUDE.md § "Testing"), so every automatable row above is uncovered — `pnpm build` is the standing stand-in, and `./scripts/vet.sh` is green. The ESLint rule files are the obvious next candidates: a `RuleTester` case beside each rule is the natural shape, and `pnpm test` already has a runner for it.

https://claude.ai/code/session_019T1QNACMv57k9bstbWSnRP


vzakharov:
Proposed squash title/body:

```
feat: adopt a strict linting ruleset (pr #16)
```

```
The lint config was thirty lines of Next presets with Prettier
wired both inside ESLint and beside it, leaving the type-aware
@typescript-eslint rules, the full jsx-a11y set and every
React-19 rule unenabled. At this size an explicit ruleset is one
pass rather than a migration, so it lands whole.

eslint.config.ts, loaded via jiti, orchestrates the preset
spreads and plugin registration; the rules live in
eslint/rule-groups/*.ts by plugin family, every one error or off
— never warn — each off carrying its rationale and, where it
delegates to the compiler, one of the twelve strictness flags
tsconfig.json gains here. Nine project-local rules land under a
vova/ namespace; ones premised on a database, server actions,
Mantine or Zod stay out, having nothing to bite on in a static
export. eslint-plugin-prettier goes, since format:check was
always the check doing the work.

What it surfaced was mostly real: an English aria-label on the
Russian CV, {stars && …} rendering a bare 0, three untitled
Spotify embeds. next-intl.d.ts registers the message catalogs
with next-intl, so useTranslations keys are checked against
en.json and useMessages() returns the catalog's real shape —
which also retired eighteen no-unsafe-* reports by replacing
t.raw(). @steiger/toolkit becomes a devDependency because the
steiger plugin's shipped declarations import it, and without it
every export it types reads as any.

Exemptions sit at the point of use — an inline disable carrying
its rationale, where whoever edits the line will read it —
rather than soften a rule to warn; a config block is for an
exemption spanning files, which leaves one, for eslint/ itself.
Where a rule's premise is simply wider than the API, the rule
learns the API instead: no-floating-promises knows node:test's
describe and it, whose promise the runner awaits.
.claude/rules/eslint.md records that and the error-or-off
policy.

```

---

docs/remove-before-merging/squash-message.md
vzakharov, line 1: cut in ~half

eslint/from-entries.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * `Object.fromEntries` that keeps the key union literal instead of widening it
+ * to `string`, so `withSeverity('error', ['a', 'b'])` types as
+ * `Record<'a' | 'b', 'error'>` and a typo in a rule name is a type error rather
+ * than a silently-inert config entry.
+ *
+ * Lives here rather than `lib/`: jiti gives `eslint/` no `@/` alias.
```
vzakharov: this comment seems out of place in this repo (unlike the source one)

messages/en.json
```diff
@@ -223,5 +223,10 @@
       "printFooter": "Full portfolio:",
       "website": "vovazakharov.com"
     }
+  },
+  "ui": {
+    "toggleTheme": "Toggle theme",
+    "switchToEnglish": "Switch to English",
+    "switchToRussian": "Switch to Russian"
```
vzakharov: why can't we just have switchToOther with only Russian/English for English/Russian respectively. That'd also clear the need for the fork at the consumption site. (It's highly improbable we'd have more languages.)

CLAUDE.md
```diff
@@ -48,12 +49,13 @@ pnpm format:check     # prettier --check .               │
 pnpm type-overlap     # scripts/type-overlap-check.ts    ┘
 ```
 
-Four things about that list are deliberate:
+Five things about that list are deliberate:
 
 - **`pnpm build` stands in for a test suite.** There isn't one (see "Testing"), so the static-export build is what catches a broken page, route or import. It is also exactly what CI runs on `main`, so a green vet means a green deploy.
 - **Never call `pnpm lint` from vet.** That script is `eslint . --fix`, which rewrites the working tree — vetting must stay read-only. `pnpm exec eslint .` is the checking form.
+- **The ESLint config is TypeScript** (`eslint.config.ts`, loaded via `jiti`), and `eslint/` is itself linted and type-checked — a broken rule file fails the lint run and the type check, not just the rules it implements. Its severity policy lives in `.claude/rules/eslint.md`.
```
vzakharov: why do we need it in claude.md if it fails anyway if uncomplied with

eslint/from-entries.ts
vzakharov: Dropped — the placement needs no defending here. 7cbb131

---

CLAUDE.md
vzakharov: Right, it's self-enforcing — a broken rule file fails both the lint run and the type check on its own. And the severity-policy pointer it carried is already in `eslint.config.ts`'s header and in `.claude/rules/eslint.md`, which is path-scoped to `eslint.config.ts` and `eslint/**` so it loads exactly when relevant. Bullet removed, back to four. 7cbb131

---

messages/en.json
vzakharov: Done. The label renders from the *current* locale's catalog, so "the other language" is a constant per file — `switchToOther` is "Switch to Russian" in `en.json` and "Переключить на английский" in `ru.json`. The consumption site is now just `t('switchToOther')`; the ternary only existed to pick between two keys that could never both apply. 7cbb131

---

docs/remove-before-merging/squash-message.md
vzakharov: Cut from ~47 lines to ~26 — three paragraphs instead of five, with the strictness-flags point folded into the rules paragraph and the two scoped `off`s into the last one. The PR comment is re-synced to match. 7cbb131

---

vzakharov:
### Attestation

Verified `2c3728c` (`claude/adopt-linting-rules-tis5zw`) against `main` @ `7058957` "chore: #17 flag repeated base combinations in type-overlap (pr #18)".

- `./scripts/vet.sh` — **green** (`pnpm build`, then `typecheck`, `eslint .`, `format:check`, `lint:fsd`, `type-overlap`, `test` in parallel), run on `f86bd07`; the two commits since it touch only the swept working tree under `docs/`, no code path
- CI on this PR — **nothing runs on a PR in this repo** (`.github/workflows/deploy.yml` triggers on `push` to `main` and `workflow_dispatch` only)
- CI-only buckets — **not dispatched: `/test-on-gh` is not part of this project's skill set**

`main` is fully contained in this branch (`/check-merge`: `target=main tip=7058957 contained`), so the base merge was a no-op. `docs/plans/` and `docs/remove-before-merging/` were swept; the branch's net diff under `docs/` is empty.

---

PR #18, opened 2026-09-01
feat: flag repeated base combinations in type-overlap

### LINKED ISSUE #17: type-overlap: flag repeated base combinations, not just repeated members
`scripts/type-overlap-check.ts` counts only a type's **own** members: the named
constituents of an intersection are treated as inherited, which is precisely
what makes `& Base` clear a finding. The blind spot is one level up — the same
*combination* of bases, repeated.

PR #11 shipped exactly that:

```ts
type ProjectCardProps = Titled & Described & { techStack?: string; … };
type ArticleCardProps = Titled & Described & { url: string };
```

Both types are clean by the current gate — neither declares an own member the
other does — yet `Titled & Described` is duplicated in the same sense a member
would be: two things to change, and the compiler catches neither the drift nor
the duplication. It was caught in review, by eye, and fixed by naming the
combination (`type Summarized = Titled & Described`). Nothing stops the next one
landing green.

## What to add

A second pairwise pass over **named constituent sets**, mirroring the existing
member pass: for every pair of type aliases, intersect the sets of type
references each lists in its intersection, and report the pair when the shared
set is large enough. The fix a finding asks for is the same shape as today's —
name the combination and intersect *that* — so the "How to fix" block and the
naming families in `scripts/type-overlap-check.README.md` carry over unchanged.

Open questions for whoever picks this up:

- **Threshold.** The member pass floors at 1; the constituent pass cannot — one
  shared base is reuse working as intended, and is the end state every member
  finding is fixed into. The natural floor is 2. Whether that is its own
  `TYPE_OVERLAP_BASES_MIN` or a fixed constant is a judgment call.
- **Transitivity.** `Summarized = Titled & Described` means a type spelling
  `Summarized` and one spelling `Titled & Described` share a combination without
  sharing a token. Resolving constituents through the alias graph catches that;
  a purely syntactic pass does not. Syntactic is consistent with the rest of the
  detector and is probably the right first cut.
- **Reporting.** Findings should group the way member findings do (one group per
  identical shared set, so one extraction clears it), which suggests factoring
  the existing pair→group collapse rather than duplicating it.

## Verification

Per `scripts/type-overlap-check.README.md` §6.9, assert the failure mode as well
as the success one: a throwaway pair of types both spelling `A & B` must exit 1
and name both locators, and must go clean once the combination is named.

Sweep at the time of PR #11: `Titled & Described` in `components/Card.tsx` was
the repo's only repeated combination, so the gate would land green today.

---

vzakharov:
## Summary

- **Adds a second pass to `pnpm type-overlap`, over each alias's named constituents.** The member pass counts only a type's *own* members — constituents are inherited, which is what lets `& Base` clear a finding — so the same *combination* spelled across two types was invisible. `Titled & Described` written twice is duplication in exactly the sense a shared member is, one level up, and the run landed green on it. Closes the gap the README named as a known blind spot.
- **Floor 2 for combinations (`TYPE_OVERLAP_BASES_MIN`), against 1 for members.** One shared base is reuse working as intended — it is the end state every member finding is fixed *into* — so 2 is the floor rather than a rung to ratchet down. Both env vars override upward only, via a shared `readThreshold`.
- **A pure combination alias now participates.** `walk` records an alias with no own members but at least one named constituent, so a type respelling `Titled & Described` is flagged against the existing `Summarized` and the fix the finding names is "use the alias that already exists". Such a decl contributes an empty set to the member pass, which is unchanged.
- **One descriptor per pass drives everything.** `findOverlaps` does the pairwise scan and pair→group collapse for either selector; `renderSection` renders a section from a heading, a noun and how a shared set prints. A `PASSES` table states each pass's threshold and wording once, so the scan, the section, the fix bullet and the clean line cannot drift apart. The run exits 1 when either pass fires.
- **The gate now has tests — the repo's first.** `pnpm test` is Node's built-in runner over `**/*.test.ts` through `tsx`: no framework, no config, wired into `vet.sh` beside the other read-only checks. Each case materializes a throwaway source tree in the OS temp directory and runs the real script against it with `cwd` set there — the script already roots its scan at `process.cwd()`, so what is asserted is the artifact `pnpm type-overlap` runs and production code gains no seam for the test's benefit. Fixtures are written at runtime because a committed `.ts` fixture that trips the gate would be scanned by the repo's own run.
- **Docs**: the README headline, §1 (why the floors differ), §2 (mechanism) and §8 (the residual — constituents match by text, so an alias and its expansion don't) plus both CLAUDE.md bullets now state the combination rule beside the member rule. §6.9 and CLAUDE.md → Testing describe the suite in place of the hand-probe recipe they carried, keeping the reason a probe cannot live in `tmp/`.

Transitivity is deliberately out of scope: constituent identity stays syntactic, consistent with the member pass, and the residual is a named gap rather than silence.

## QA Checklist

Every row below now runs as `pnpm test`, in `scripts/type-overlap-check.test.ts` — checked off because the suite asserts it, not because someone looked once.

- [x] `clean-run` — on a clean tree, `pnpm type-overlap` exits 0 and prints `type-overlap: clean — no shared members (min=1), no repeated base combinations (min=2).`
- [x] `base-probe` — two types both spelling `Titled & Described` produce exit 1 and a `== Repeated base combinations (threshold=2) ==` section naming both, with an existing pure-combination alias pulled into the same group.
- [x] `alias-reuse` — replacing the two spellings with one named combination both types intersect takes the run clean again.
- [x] `member-probe` — two types sharing one member still fire the `== Overlapping type shapes (threshold=1) ==` section, and a named constituent is still counted as inherited.
- [x] `both-fire` — a tree tripping both passes prints both sections *and* both sub-bullets under fix step 1; one tripping a single pass prints only that pass's sub-bullet.
- [x] `env-floors` — `TYPE_OVERLAP_BASES_MIN=1` and `TYPE_OVERLAP_MIN=nope` each fail with a message naming the var and its floor; `TYPE_OVERLAP_MIN=2` suppresses a single-shared-member finding.
- [x] `skips` — a duplicate under `tmp/`, in a `*.test.ts`, in a `*.d.ts`, in a dot-directory or under `public/` is not scanned.
- [x] `vet` — `./scripts/vet.sh` is green on the branch, base merged.

| `clean-run` | integration | ✅ | `type-overlap: clean runs` — asserts the clean line verbatim |
| `base-probe` | integration | ✅ | `the base-combination pass` — exit code, section, group listing, the alias pulled in |
| `alias-reuse` | integration | ✅ | `goes clean once both types reuse the named combination` |
| `member-probe` | integration | ✅ | `the member pass` — regression guard on the pre-existing pass |
| `both-fire` | integration | ✅ | `the fix bullets` — sub-bullets emitted per firing pass, not unconditionally |
| `env-floors` | integration | ✅ | `thresholds` — upward override accepted, floor and non-integer refused |
| `skips` | integration | ✅ | `what the scan skips` — one case per skip rule |
| `vet` | e2e | ✅ | `./scripts/vet.sh`, which now runs `pnpm test` alongside the other read-only checks |

Each case was checked against a deliberately mutated script to confirm it fails when the behavior it names is removed — disabling constituent collection reddens 3, dropping pure-combination aliases reddens exactly the 1 that guards it, lowering the base floor reddens 13, un-skipping `tmp/` reddens 1.

Merging the base pulled in the content pipeline, and the new pass had a finding waiting for it: `RenderedDocument` and `ArticleHeaderProps` both spelled `Titled & WithReadingMinutes`. That pair is now `Headlined` in `lib/content/render.ts`.

Closes #17

https://claude.ai/code/session_01SiVzXKtvx6sPdryPG9sNNp


vzakharov:
Proposed squash title/body:

```
chore: #17 flag repeated base combinations in type-overlap (pr #18)
```

```
The gate counted only a type's own members, so the named constituents
of an intersection were invisible to it: `Titled & Described` spelled
across two types is one shape under two spellings, one level up, and
the run landed green on it. A second pairwise pass now compares those
constituents, at a floor of 2 (`TYPE_OVERLAP_BASES_MIN`) where members
sit at 1 — one shared base is reuse working as intended, and is the end
state every member finding is fixed into. A pure combination alias
participates despite having no own members, so the fix a finding names
is "use the alias that already exists".

Both passes share one pairwise scan and group collapse, one section
renderer, and a `PASSES` table stating each pass's threshold and
wording once, so the scan, the section and the fix bullet cannot drift
apart. Constituent identity stays syntactic, matching the member pass,
which leaves an alias and its expansion unmatched; the README carries
that residual in place of the gap this closes.

The gate also gains the repo's first tests. `pnpm test` is Node's
built-in runner over `**/*.test.ts` through `tsx`, wired into `vet.sh`;
each case materializes a throwaway source tree in the OS temp directory
and runs the real script against it, so what is asserted is the
artifact `pnpm type-overlap` runs and production code gains no seam for
the test's benefit. A committed fixture would be a `.ts` file the gate
itself then scanned.

The new pass had a finding waiting: `Titled & WithReadingMinutes`,
spelled in both `RenderedDocument` and `ArticleHeaderProps`, is now
`Headlined` in `src/shared/content/render.ts`.

Closes #17

```

---

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,42 @@
+Proposed squash title/body:
+
+```
+feat: #17 flag repeated base combinations in type-overlap (pr #18)
```
vzakharov: not a feat

scripts/type-overlap-check.ts
vzakharov, line 1: did we do some tests to make sure this actually works as intended?

docs/remove-before-merging/squash-message.md
vzakharov: Fixed — `chore:` now. Tooling change, nothing user-facing ships.

---

scripts/type-overlap-check.ts
vzakharov: Only by hand, which wasn't good enough — so there are real tests now, in `b25f885`. `scripts/type-overlap-check.test.ts`, 18 cases on Node's built-in runner (`pnpm test`, wired into `vet.sh`). No framework added.

Each case writes a throwaway source tree into the OS temp dir and runs the actual script against it with `cwd` pointed there, so it exercises the same entrypoint `pnpm type-overlap` does — no test seam in production code. It asserts both directions: the clean line verbatim, and per pass the section heading, the group listing and the fix sub-bullet; plus the threshold overrides and every skip rule.

To check the tests aren't vacuous I ran them against a deliberately broken script: disabling constituent collection reddens 3, dropping pure-combination aliases reddens exactly the 1 that guards it, lowering the base floor to 1 reddens 13, un-skipping `tmp/` reddens 1.

It also caught something real. Merging `main` brought in the content pipeline, and the new pass fired on it — `RenderedDocument` and `ArticleHeaderProps` both spelled `Titled & WithReadingMinutes`. That's now `Headlined` in `lib/content/render.ts`. Vet is green on the merged base.

---

docs/remove-before-merging/squash-message.md
vzakharov: cut in ~half

vzakharov:
### Attestation

Verified `916681f` (`claude/type-overlap-base-combinations-v1vnm8`) against `main` @ `bd0087c` "refactor: adopt Feature-Sliced Design, enforced by Steiger (pr #12)".

- `./scripts/vet.sh` — **green** (build, typecheck, eslint, format:check, lint:fsd, type-overlap, and the new `pnpm test`)
- CI on this PR — **nothing runs on a PR in this repo** (`deploy.yml` triggers on push to `main` and `workflow_dispatch` only)
- CI-only buckets — **not dispatched: this project has no `/test-on-gh` lane**, and the diff reaches no surface needing real credentials, services or a browser

The vet run predates only the working-artifact sweeps (`docs/issue/`, `docs/plans/`, `docs/remove-before-merging/`), which touch no code path; the code state attested is the branch's current head.

---

PR #19, opened 2026-09-01
feat: replace Tailwind with Mantine as the styling layer

vzakharov:
## Summary

- **Replaces Tailwind CSS 4 with Mantine 9** as the styling and component layer. `tailwindcss`, `@tailwindcss/postcss` and `next-themes` are gone; Mantine now owns the colour scheme, covering the same light/dark/system toggle one for one. The static export is untouched — Mantine is CSS-variable and CSS-module based, so nothing injects styles at runtime.
- **The look is preserved exactly, and that was checked rather than assumed.** Every page was rendered before and after and compared numerically (computed font sizes, line heights, opacity and element geometry, via CDP) on `/`, `/en/cv` and `/ru/cv`, at 768px and 1280px, in both schemes, and in print media. Screen geometry matches element-for-element; the print rendering matches the old type scale and lands on the same four pages. Two deliberate deviations are listed below.
- **The structure is the one Mantine's own guide lays out**, with the pieces this repo needs: a server `RootLayout` holding `mantineHtmlProps` + `<ColorSchemeScript>` with a client `ThemeProvider` beside it, tokens bound through a `cssVariablesResolver` rather than a `:root` block, a typed `cssColor()` helper, and co-located `.module.scss` linted by stylelint. Every stylesheet being Sass means `postcss-preset-mantine`'s mixins are unreachable, so `styles/_mantine.scss` carries Sass counterparts and the repo keeps no PostCSS config at all.
- **`CVPage.tsx` and `HomePage.tsx` are split** — `CvSection`, `CvSubsection`, `CvBullets`, `ExperienceEntry`; `DevSection`, `MusicSection`, `WritingSection`, `ContactSection` over a shared `Section`/`Subheading` pair — along seams the conversion exposed, not a speculative refactor. `hooks/useMounted.ts` gives way to the one `@mantine/hooks` ships, and `vet.sh` gains stylelint as a fourth concurrent check.

### Deliberate deviations

- **The CV's theme and locale buttons are 38×38 rather than 38×46 and 40×46.** The old height came from `align-items: stretch` on the header row combined with a `text-lg` class, and it differed between the two pages for no design reason; matching it would mean threading a height prop through two shared components for one call site.
- **The printed CV starts ~30px higher on page 1.** Tailwind's `space-y-8` put a margin above the header because the `display: none` print-button row still counted as a sibling; a Stack's gap does not. Page breaks are otherwise identical.

### Open question from the plan, taken as recommended

`lucide-react` stays rather than moving the glyphs to `public/` SVGs — that pattern earns itself across a large icon set, and this site has four. Say the word if you'd rather have it the other way; it's a separable change.

## QA Checklist

- [ ] `home-light` — open `/` in light mode: pure white ground, black text, square hairline card borders, Merriweather body, JetBrains Mono tech-stack lines.
- [ ] `home-dark` — same page in dark mode: the palette inverts to pure black/white with no third colour anywhere.
- [ ] `theme-cycle` — click the toggle: light → dark → system, with no flash of the wrong scheme on reload.
- [ ] `cv-en` — open `/en/cv`: heading scale, card borders and dimmed period/tech lines match the pre-change page.
- [ ] `cv-ru` — open `/ru/cv` and round-trip the locale toggle; the longer Russian copy wraps without overflow.
- [ ] `cv-print` — hit the `/ PDF` button and print to PDF: A4, header controls hidden, the print-only footer shown, dimming flattened to full black, cards reduced to a rule between entries, four pages.
- [ ] `nav-links` — nav links underline on hover only; prose, contact and footer links are underlined at rest.
- [ ] `hover-states` — card borders darken on hover; the theme, locale and print buttons invert to solid foreground.
- [ ] `tablet` — view `/` and `/en/cv` at ~700px: the page padding steps up at 640px and the project/article grids go two-up at 768px.
- [ ] `no-tailwind` — `grep -ri tailwind` over the tree returns nothing outside `pnpm-lock.yaml`.

| Item           | Automatable | Covered? | Notes                                                                                       |
| -------------- | ----------- | -------- | ------------------------------------------------------------------------------------------- |
| `home-light`   | manual-only | —        | Visual match against the pre-change render; no reference-image harness exists                |
| `home-dark`    | manual-only | —        | As above, in the dark scheme                                                                 |
| `theme-cycle`  | e2e         | ❌       | Drive the toggle, assert `data-mantine-color-scheme` cycles and survives a reload            |
| `cv-en`        | manual-only | —        | Visual match                                                                                 |
| `cv-ru`        | e2e         | ❌       | Load `/ru/cv`, click the locale toggle, assert the route and copy switch                     |
| `cv-print`     | manual-only | —        | Needs a human eye on the paginated PDF; the print media query is not assertable in the build |
| `nav-links`    | e2e         | ❌       | Assert `text-decoration-line` per link group in the rendered page                            |
| `hover-states` | manual-only | —        | Hover appearance is a judgement call against the old design                                  |
| `tablet`       | e2e         | ❌       | Render at 700px/800px and assert the padding and grid-column counts                          |
| `no-tailwind`  | unit        | ❌       | A one-line repo grep, cheap to wire into `vet.sh` if it is worth guarding                    |

There is no test suite yet (CLAUDE.md → Testing), so every automatable row above is uncovered. `pnpm build` and the vet suite are green; the visual rows were verified by rendering both revisions side by side, but nothing in CI protects them.

https://claude.ai/code/session_01MCHce8FTH5sVnx9y7LPcKp

---

### Review round 1 (`668ba0a`)

Every comment on the first review is addressed in that commit; each thread carries the specifics. In short: the PostCSS config is gone entirely (both plugins were dead — the built CSS is byte-identical without them), the global sheets are `.scss`, the colour tokens sit behind a Sass mixin, `!important` is a stylelint error, `Providers` is `ThemeProvider`, `HomePage` is split four ways, `Suffixed<Base, Suffix>` lands in `shared/typings`, and the squash message is ~35% shorter with the archaeology swept out.

Two claims in the original body no longer hold and are corrected above: the structure is Mantine's documented one rather than a house port, and there is no PostCSS config to speak of.


vzakharov:
Proposed squash title/body:

```
feat: swap Tailwind for Mantine as the styling layer (pr #19)
```

```
Tailwind's footprint here was small and fully enumerable — six files
using className, one PostCSS plugin, one globals.css — so this is a
change of mechanism, not a restyle. The visual identity is preserved,
and that was checked rather than assumed: both revisions were rendered
and their computed typography, opacity and geometry compared across
every page, both colour schemes, two widths and print media.

Mantine 9 takes the colour scheme over from next-themes, covering the
same light/dark/system toggle with one provider boundary instead of
two. Colour tokens live in globals.scss as --color-* and reach TSX
through a typed cssColor(); Mantine's own variables bind to them in a
cssVariablesResolver, because it renders its variable block into <body>
where a stylesheet's :root cannot outrank it.

Two cascade facts decide where every rule lands, both recorded in
.claude/rules/styling.md. Mantine ships layered, so unlayered rules beat
it and !important is now a stylelint error. Style props, varsResolver
output and visual defaultProps render inline, out of reach of any
stylesheet — so print rules split between print.scss and the components
that own them, and the control skin lives in variantColorResolver.
Every stylesheet being Sass leaves Mantine's PostCSS plugins nothing to
do and the repo carries no PostCSS config at all, which leaves two
scales Sass and TypeScript both need and neither can read across — the
breakpoints and the colour-token names. Both are written once in
TypeScript and their Sass halves generated by pnpm styles:codegen,
which vet runs and fails on whenever it had to rewrite one.

Rendered markdown keeps a scale of its own. Mantine's Typography is
sized for controls — 1.55 leading, a fixed 12px on code, one 10px gap
under every heading, all off the same absolute ladder as buttons and
inputs — so prose.scss states the long-form rhythm outright, in em, and
the markup is wrapped in nothing. CVPage and HomePage split along the
seams the conversion exposed, which moves the root page's English-copy
lint exemption from one file to the whole slice; useMounted gives way
to the @mantine/hooks one.

```

---

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,47 @@
...
+
+```
+feat: swap Tailwind for Mantine as the styling layer (pr #19)
+```
+
+```
+Tailwind's footprint here was small and fully enumerable — six files
+using className, one PostCSS plugin, one globals.css — which made it
+cheap to move the site onto Mantine and align it with the conventions
+Playgramai/playgramapp already runs on. Mantine is the mechanism, not a
```
vzakharov: mentioning playgram is archaeology; sweep for that

docs/remove-before-merging/squash-message.md
vzakharov, line 1: cut by 30-50%

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,47 @@
...
+it for MantineProvider. Colour tokens live in globals.css as --color-*
+and reach TSX through a typed cssColor(); Mantine's own variables are
+bound to them through a cssVariablesResolver rather than a :root block,
+because the provider injects its stylesheet at runtime and outranks
+one. Component CSS is co-located .module.scss, linted by a stylelint
+run that joins the vet suite as a fourth concurrent check.
+
+Two cascade facts shape where every rule had to go, and .claude/rules/
+styling.md records them. Mantine ships as styles.layer.css, so any
+unlayered rule beats it and nothing in this tree needs !important. But
```
vzakharov: can we prohibit !important via stylelint while we're at it?

src/app/styles/globals.css
vzakharov, line 1: why not scss?

src/app/styles/globals.css
```diff
@@ -1,199 +1,17 @@
...
 
+/* Colour tokens. `src/shared/ui/css-color.ts` mirrors this list as a union type
+   and is how TSX reaches them; the two must stay in step. */
 :root {
-  --background: #ffffff;
-  --foreground: #000000;
+  --color-background: #fff;
+  --color-foreground: #000;
+  --color-border-hairline: rgb(0 0 0 / 20%);
+  --color-border-hairline-strong: rgb(0 0 0 / 40%);
```
vzakharov: define via mixin (I see two uses here and more in print.scss)?

src/app/styles/print.scss
```diff
@@ -0,0 +1,109 @@
...
+  h4,
+  h5,
+  h6 {
+    break-after: avoid;
+  }
+
+  h1 {
+    font-size: 20pt;
+    line-height: 1.2;
+    margin-bottom: 0.2cm;
```
vzakharov, line 54: mixin?

src/app/styles/prose.scss
vzakharov, line 1: why not scss?

src/app/styles/theme.ts
```diff
@@ -0,0 +1,106 @@
...
+  // Sizes live here rather than on call sites so that `print.scss` can re-key
+  // them off `h1`–`h4`: a `fz` prop would render as an inline style no print
+  // rule could reach.
+  headings: {
+    fontFamily: 'var(--font-merriweather), serif',
+    sizes: {
+      h1: { fontSize: '36px', lineHeight: '40px' },
+      h2: { fontSize: '30px', lineHeight: '36px' },
+      h3: { fontSize: '24px', lineHeight: '32px' },
+      h4: { fontSize: '20px', lineHeight: '28px' },
```
vzakharov, line 56: why are we defining this here and print there? for this specifically and for the theme generally, is there anything that would belong better to the (s)css file?

src/app/ui/theme-provider.tsx
vzakharov, line 1: I don't like the name, seems to talk about the mechanism than purpose. (I'd understand the need if we had more than one provider, but we don't seem to atm.)

src/app/ui/providers.tsx
```diff
@@ -0,0 +1,32 @@
+'use client';
+
+import '@mantine/core/styles.layer.css';
+import { MantineProvider, type CSSVariablesResolver } from '@mantine/core';
+import { cssColor } from '@/shared/ui';
+import { theme } from '../styles/theme';
+
+// Must go through the resolver rather than a plain `:root` block in
+// globals.css: Mantine's stylesheet ships after it under Turbopack and wins a
+// plain override.
```
vzakharov: Why wasn't this a problem in the playgram repo?

src/app/ui/root-layout.tsx
```diff
@@ -31,15 +33,21 @@ export function RootLayout({
...
+    <html
+      lang="en"
+      className={`${merriweather.variable} ${jetbrainsMono.variable}`}
+      {...mantineHtmlProps}
+    >
+      <head>
+        <ColorSchemeScript defaultColorScheme="auto" />
+      </head>
+      <body>
         <NextIntlClientProvider>
```
vzakharov, line 52: ah, re. providers, we do, but it wasn't even placed in `providers.tsx` -- kinda reiterating my point

src/pages/home/ui/home-page.tsx
vzakharov, line 1: let's split to <450 lines any

src/shared/ui/css-color.ts
```diff
@@ -0,0 +1,15 @@
...
+ * The site's colour tokens, declared in `src/app/styles/globals.css` as
+ * `--color-*` properties and redeclared there for the dark scheme, so call
+ * sites never branch on the scheme. A token missing from this union is a type
+ * error rather than a silently dead `var()`.
+ */
+export type CssColor =
+  | 'background'
+  | 'foreground'
+  | 'border-hairline'
+  | 'border-hairline-strong';
```
vzakharov: define and use:

```
type Suffixed<Base extends string, Suffix extends string | number> = `${Base}${
  | (Suffix extends '' ? '' : never)
  | `-${Suffix}`}`;
```

src/shared/ui/internal-link.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * polymorphic `component` prop cannot take it from a server component. The
+ * pairing lives behind this client boundary instead.
+ */
+export function InternalLink({
+  href,
+  children,
+  ...props
+}: AnchorProps & { href: string; children: ReactNode }) {
+  return (
+    <Anchor component={Link} href={href} {...props}>
```
vzakharov: Why are we doing this instead of `<Link ...`? (Not insisting, asking for now)

postcss.config.cjs
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  plugins: {
+    'postcss-preset-mantine': {},
+    'postcss-simple-vars': {
+      // Kept in step with `theme.breakpoints` in src/app/styles/theme.ts and
+      // the `$breakpoint-*` variables in src/shared/styles/_mantine.scss.
```
vzakharov: I'm lost (this is not the only offending line ofc): why do we need all three? That's not an objection for now -- just a request to explain to to me clearly, maybe through some (real or hypothetical) example

postcss.config.cjs
vzakharov, line 1: why not .ts

docs/remove-before-merging/squash-message.md
vzakharov: Swept. Gone from the squash message and from the PR body, which claimed the structure was a port of house conventions — it is Mantine's own documented layout, so that is what it says now.

The one place I left it is `docs/plans/migrate-mantine-tailwind.completed.md`: comparing against a repo we already run is what the plan was *for*, and `/finalize` deletes the whole `docs/plans/` tree before the squash, so none of it reaches `main`. Say the word if you'd rather it went now.

(`scripts/type-overlap-check.README.md` also mentions it, but that line is on `main` already and predates this branch — out of scope here.) — 668ba0a

---

docs/remove-before-merging/squash-message.md
vzakharov: Done — 33 body lines to 22, about 35%. Rewritten rather than trimmed, since three of the facts it stated also changed in this round. — 668ba0a

---

docs/remove-before-merging/squash-message.md
vzakharov: Yes — `declaration-no-important: true` is on. The tree had none, so it is a ratchet rather than a cleanup, and the rule comment says why one could only ever be fighting another rule of ours: Mantine ships layered, so every unlayered rule here already outranks it whatever the specificity. — 668ba0a

---

src/app/styles/globals.css
vzakharov: No good reason — it is `globals.scss` now, and `prose.scss` with it. The original split ("global sheets plain `.scss`, but these two `.css`") was a leftover from the Tailwind file it replaced, not a decision. Making them Sass is also what lets the token mixin below work. — 668ba0a

---

src/app/styles/globals.css
vzakharov: Done. `styles/_tokens.scss`:

```scss
@mixin colors($background, $foreground, $hairline, $hairline-strong) { … }
```

Three call sites now — light and dark in `globals.scss`, print in `print.scss` — so the four property names are written once instead of twelve times. — 668ba0a

---

src/app/styles/print.scss
vzakharov: Same mixin, third call site: `@include tokens.colors(#fff, #000, #ddd, #ddd);`.

The rest of `print.scss` I left as literal element rules. They are not three restatements of one shape — each `h1`–`h4` block carries its own size, line height and margin — so a mixin there would be a four-argument wrapper around a four-line body. — 668ba0a

---

src/app/styles/prose.scss
vzakharov: Same as `globals.css` — `prose.scss` now, with the block comments converted to `//` so they stop being emitted into the output. — 668ba0a

---

src/app/styles/theme.ts
vzakharov: Fair question, and the split is forced from both ends rather than chosen.

**Why the sizes cannot be in CSS.** `theme.headings.sizes` compiles to CSS variables — the built page carries `--mantine-h1-font-size: 36px`, `--mantine-h1-line-height: 40px` and so on — so on the face of it a `:root` block would do. It does not, because of *where* Mantine puts that block. It renders as `<style data-mantine-styles>` at the **top of `<body>`**, i.e. after every stylesheet in `<head>`, and on the same `:root` selector. Equal specificity, later in the document, Mantine wins. That is the same reason `cssVariablesResolver` exists rather than a `:root` override of `--mantine-color-body`, and I have rewritten that comment too — it previously blamed Turbopack, which was wrong (see the thread on `providers.tsx`).

**Why print cannot be in the theme.** A Mantine theme has no media-query axis: `headings.sizes` is one set of values, full stop. So the print scale has to be a stylesheet rule, and it keys off `h1`–`h4` — which only works because the sizes came from the theme as element-level variables rather than from `fz` props, which would have rendered as inline styles no rule could reach.

So: **screen values in `theme.ts` because it is the only place that outranks Mantine; print values in `print.scss` because it is the only place that has `@media print`.**

**Anything that would move?** Of the theme, nothing — every entry is either a Mantine variable (loses in CSS, per above) or a `classNames`/resolver binding, which is not a value at all. Going the other way, what is already in CSS stays there for the mirror-image reason: `--color-*` is safe in `globals.scss` precisely because Mantine never declares those names, so nothing is racing it.

`.claude/rules/styling.md` now states the mechanism rather than the folklore. — 668ba0a

---

src/app/ui/theme-provider.tsx
vzakharov: Agreed — renamed to `theme-provider.tsx`, exporting `ThemeProvider`. It is the file `components/ThemeProvider.tsx` used to be before this branch deleted it, so the name is not new here either.

See the sibling thread on `root-layout.tsx` for the second half of your point. — 668ba0a

---

src/app/ui/root-layout.tsx
vzakharov: You are right that there are two, and I went the other way on the fix: `NextIntlClientProvider` stays in `root-layout.tsx`.

The reason is that it is not ours to collect. It is next-intl's own component, used directly with no configuration — nothing about it needs a boundary of our own, so pulling it into a bag would add a file layer whose only content is a re-export. What the Mantine one wraps, by contrast, is a `cssVariablesResolver`: a function prop, which is not serialisable across the server boundary, so a client component genuinely has to exist for it.

So there is no "providers" bag any more — there is `ThemeProvider`, which does one nameable thing, and next-intl's provider used as-is. That is what you were pointing at, I think: the collection was never earning the generic name. — 668ba0a

---

src/app/ui/providers.tsx
vzakharov: It was not a Turbopack thing at all — that comment was repeating a claim I had taken on trust, and it was wrong. I checked the built output.

Mantine renders its variable block as `<style data-mantine-styles>` at the **top of `<body>`**, on `:root, :host`. Every stylesheet is a `<link>` in `<head>`. Same specificity, later in the document, so a `--mantine-color-body` written in `globals.scss` loses — under any bundler, in any framework. Nothing repo-specific about it, and nothing to have been "a problem" elsewhere: any project overriding a `--mantine-*` variable hits the same wall, which is why the resolver is Mantine's documented answer.

Worth separating from the case that *does* work: `--color-*` lives in `globals.scss` and is fine there, because Mantine never declares those names — no race to lose. The comment and `.claude/rules/styling.md` now both say this. — 668ba0a

---

src/pages/home/ui/home-page.tsx
vzakharov: Split — 522 lines to 96. Four section files (`dev-section`, `music-section`, `writing-section`, `contact-section`, the largest 193 lines) over a shared `section.tsx` holding `Section` and a new `Subheading`, which absorbs the `<Title order={3} mt={SUBHEADING_GAP}>` that was repeated six times. `home-page.tsx` keeps the shell: hero, nav, the four sections, footer. `SpotifyEmbed` moved to the only section that uses it, and `home-page.module.scss` followed its single consumer to `dev-section.module.scss`.

`DevSection` is `async` and does the `renderPrimaryDocuments` await itself, so `HomePage` is a plain sync component now.

Verified rather than assumed: rendered markup across all twelve pages is byte-identical (diffing the built HTML with chunk hashes and CSS-module hashes normalised), and the built CSS rule set matches too — the only textual change is the module filename inside the generated class name. — 668ba0a

---

src/shared/ui/css-color.ts
vzakharov: Defined and used, with one change to the definition — the version in your comment admits a trailing hyphen:

```ts
// Suffix = '' | 'strong' distributes, and `-${''}` is '-', so the union is
// 'border-hairline' | 'border-hairline-' | 'border-hairline-strong'.
```

A conditional gets the intent without the stray member:

```ts
export type Suffixed<Base extends string, Suffix extends string | number> =
  Suffix extends '' ? Base : `${Base}-${Suffix}`;
```

Lives in `src/shared/typings`, used twice here:

```ts
export type CssColor =
  | 'background'
  | 'foreground'
  | Suffixed<'border-hairline', '' | 'strong'>;

export function cssColor(color: CssColor): `var(--${Suffixed<'color', CssColor>})` { … }
```

The return annotation is the second one — it makes `cssColor()` yield a literal type rather than `string`, which a template expression would otherwise widen to. — 668ba0a

---

src/shared/ui/internal-link.tsx
vzakharov: Because a bare `<Link>` renders an unstyled `<a>` — no site link colour, no `underline` default, none of the `size`/`inherit`/`c` props the call sites use. `Anchor` is what carries all of that, and Mantine's way to make an `Anchor` route internally is `component={Link}`.

The wrapper exists only because that prop cannot cross the server boundary: `component` takes the `Link` *component itself*, and React refuses to serialise a component reference from a server component to a client one. So something with `'use client'` has to hold the pairing, and this file is the smallest thing that can.

The alternative is a plain `<Link className={…}>` with the link styling restated in a CSS module — which trades the wrapper for a second source of truth for what a link looks like. Happy to go that way if you prefer it; the wrapper looked like the cheaper of the two. — 668ba0a

---

postcss.config.cjs
vzakharov: Sorry — this thread and the `.ts` one below were the two I left unanswered. Here it is properly.

**The three were three languages each needing the same number, and none can read another's declaration.**

1. **`theme.breakpoints`** (TypeScript) — Mantine reads it for responsive *props*. `<SimpleGrid cols={{ base: 1, md: 2 }} />` becomes `@media (min-width: 48em)` because `md` is `48em` there.
2. **`mantine-breakpoint-*` in `postcss.config.cjs`** (`postcss-simple-vars`) — lets a plain `.css` file write `@media (max-width: $mantine-breakpoint-sm)`; PostCSS substitutes the literal.
3. **`$breakpoint-*` in `_mantine.scss`** (Sass) — the same job as 2, for a `.scss` file.

So **2 and 3 are one job in two languages**, and you only need whichever your stylesheets are written in. Every stylesheet here is `.scss`, which made 2 dead weight — the config is gone, both plugins with it, and the built CSS came out byte-identical.

**Why 1 and 3 can't collapse into one.** A media query condition cannot read a CSS custom property, so the number has to be a literal by build time — and TypeScript and Sass share no build step here. Two is the floor. Concretely, in a `.module.scss`:

```scss
@media (min-width: $mantine-breakpoint-sm) { … }         // Sass: "Undefined variable" — build fails
@media (min-width: var(--mantine-breakpoint-sm)) { … }   // parses, and silently never matches
@include mantine.larger-than(mantine.$breakpoint-sm) { … }  // works
```

The middle one is the trap: valid CSS, no warning, just a query that never fires.

**And what "change them together" is protecting.** Set `sm: '42em'` in `theme.ts` alone, and on `/case-studies` the grid's `cols={{ base: 1, sm: 2 }}` starts going two-up at 42em while the same file's `larger-than($breakpoint-sm)` padding step still fires at 40em — two columns squeezed into the narrow padding for 2em of viewport. Nothing errors; it just looks wrong in a 32px-wide band.

One more thing on the line you flagged: it pointed at `src/shared/styles/_mantine.scss`, which has never existed — the file is root `styles/_mantine.scss`. Fixed in `e4c59f4` along with the rest of that comment.

---

postcss.config.cjs
vzakharov: Because Turbopack does not load one — and it fails *silently*, which is worse than rejecting it. I probed both, each with a config naming a nonexistent plugin:

| file | result |
| --- | --- |
| `postcss.config.ts` | build passes — the file is ignored outright |
| `postcss.config.mjs` | build fails: `Cannot find module 'postcss-probe-does-not-exist'` |

So a `.ts` config would have looked right, type-checked, and done nothing.

`.cjs` had no such excuse, though, and you're right to poke at it. The repo is `"type": "module"`, the file this branch replaced was `postcss.config.mjs`, and `.mjs` demonstrably works — switching to `.cjs` and `module.exports` was a step backwards, and I took it by copying a shape rather than checking. Moot now that there's no config at all, but for the record: if one ever comes back it should be `.mjs`.

---

postcss.config.cjs
vzakharov: Codegen, in the TS→Sass direction — `eefe24c`.

The numbers now live once, in `src/app/styles/breakpoints.ts`, and `styles/_breakpoints.scss` is generated from it by `scripts/generate-breakpoints.ts`. `_mantine.scss` `@forward`s the generated partial, so every `mantine.$breakpoint-sm` call site is untouched and the compiled CSS is identical (11 media queries, the same two widths — only `sm` and `lg` are actually reached from Sass).

TS is the source rather than Sass because it's the side that has a consumer at runtime (Mantine reads `theme.breakpoints` to resolve `cols={{ base: 1, md: 2 }}`) and the side a type can constrain — the map is `satisfies MantineBreakpointsValues`, so dropping a key is a type error rather than a missing Sass variable.

And you get the lint rule too, because codegen alone doesn't close the hole — a generated file that nothing checks is just a copy with a nicer comment on it. `pnpm styles:breakpoints --check` compares the partial to what a write would produce and joins the vet fan-out. Both drift directions fail, verified:

```
$ sed -i "s/sm: '40em'/sm: '42em'/" src/app/styles/breakpoints.ts   # edit TS, forget to regenerate
$ pnpm styles:breakpoints --check
styles/_breakpoints.scss is not what src/app/styles/breakpoints.ts generates:

  line 6
    expected: $breakpoint-sm: 42em;
    found:    $breakpoint-sm: 40em;

Run `pnpm styles:breakpoints`.
$ echo $?
1
```

…and the same output, pointing the other way, if someone hand-edits the partial instead. It reads two files and writes nothing, so it overlaps the other seven checks safely.

Worth noting for the record: this is a hole `main` had too, not something the migration introduced — the pre-migration `postcss.config.mjs` carried its own third copy of the scale. That copy is gone; this closes the remaining pair.

One pair of the same shape is still hand-kept: the `--color-*` names in `styles/_tokens.scss` and the `CssColor` union in `src/shared/ui/css-color.ts`, whose comment likewise says "the two must stay in step." It resists the same treatment because only the *names* are shared — the values differ per palette — so the fix is shaped differently: have the mixin take a Sass map and `@each` over a generated key list, which makes the call sites more verbose. Happy to do it if you want the symmetry; left alone for now as it's outside this PR.

---

src/shared/ui/css-color.ts
```diff
@@ -1,15 +1,18 @@
+import type { Suffixed } from '@/shared/typings';
```
vzakharov: we only ever use it here, and I don't imagine any other use for it, so I'd not contaminate the shared typings module

styles/_mantine.scss
```diff
@@ -1,20 +1,18 @@
...
-// `mantine-breakpoint-*` variables in postcss.config.cjs.
-$breakpoint-xs: 30em;
-$breakpoint-sm: 40em;
-$breakpoint-md: 48em;
-$breakpoint-lg: 64em;
-$breakpoint-xl: 80em;
+// The breakpoint scale, generated from src/app/styles/breakpoints.ts. Forwarded
+// rather than used, so `@use '…/_mantine' as mantine` reaches
+// `mantine.$breakpoint-sm` alongside the mixins below.
+@forward 'breakpoints';
```
vzakharov: pls explain forwarding (I'm not familiar with the concept) vs using to me, generally and in this specific case

styles/_tokens.scss
```diff
@@ -0,0 +1,10 @@
+// The site's colour tokens — one call per palette (light, dark, print).
+// `src/shared/ui/css-color.ts` mirrors the names as a union type and is how TSX
+// reaches them; the two must stay in step.
```
vzakharov: can we have a checker to make sure they don't diverge? maybe instead of writing a new script we can add it to the breakpoint checker (renaming it of course), making it a single styling codegen/checker script

src/shared/ui/css-color.ts
vzakharov: Agreed, and it is gone rather than moved — `src/shared/typings` keeps only what genuinely crosses slices, and this crossed nothing.

Removing it outright rather than relocating it is the part worth flagging, since you asked for it defined and used one round ago. Both of its uses dissolved:

- **The union** is no longer hand-written. Making the Sass side generated (see the `_tokens.scss` thread) needs the names as a *runtime* value, so `CSS_COLORS` is a `const` array and `CssColor` derives from it — CLAUDE.md’s own array-then-derive rule. A flat array spells `border-hairline` and `border-hairline-strong` out, so the family expression `Suffixed` existed for has nothing left to express.
- **The return type** was its last caller, and with no empty suffix in play `` `var(--color-${CssColor})` `` says the same thing without the helper.

So keeping it would have left a one-use alias with a dead `Suffix extends ''` branch. Say the word if you would rather have the family back as an expression and the array derived some other way — but I do not think it pays for itself at two names.

---

styles/_mantine.scss
vzakharov: Happy to — it is one of Sass's genuinely non-obvious corners.

**Generally.** Both load a module; they differ in *who* gets to see its members afterwards.

- `@use 'x'` loads `x` **for this file**. You reach its members namespaced (`x.$foo`, `x.mixin()`). It is an import, and it is **not transitive**: a third file that `@use`s *you* sees nothing of `x`.
- `@forward 'x'` makes `x`'s members part of **your own public API**, so whoever `@use`s you reaches them under *your* namespace. It is a re-export — and, symmetrically, it does **not** give you access to them yourself.

So the two are complements, not alternatives, and a file that needs both writes both lines:

```scss
@use 'x';      // I want to call x.thing()
@forward 'x';  // ...and I want my consumers to reach it too
```

(Sass loads the module once regardless; the duplication is about visibility, not work.)

**Here.** Call sites write one line, `@use '…/_mantine' as mantine`, and reach `mantine.smaller-than(…)` and `mantine.$breakpoint-sm` from it. That works because `_mantine.scss` forwards the generated partial.

`@use 'breakpoints'` there instead would invert it exactly: `_mantine.scss` could say `$breakpoint-sm` internally, and every call site would see no breakpoints at all — each would need its own second `@use '…/_breakpoints'` line, and the fact that the scale lives in a separate generated file would leak into every stylesheet in the tree.

And `@forward` alone — rather than both lines — because `_mantine.scss` has no use for the numbers itself: `smaller-than($breakpoint)` and `larger-than($breakpoint)` take the value as an argument. It is purely a pass-through, which is the case `@forward` is for. The split into a generated file is then invisible to everything downstream, which is the point: codegen owns a whole file, and `_mantine.scss` stays the one entry point.

---

styles/_tokens.scss
vzakharov: Done, and folded into the breakpoint script as you suggested — `scripts/generate-breakpoints.ts` is now `scripts/generate-styles.ts`, `pnpm styles:codegen`, covering both scales. `a731eeb`.

It generates rather than checks, because generating turned out to be available for free. A checker needs the TS names at runtime to compare against; once they exist as a runtime value, writing the file is strictly stronger than diffing it. So `CSS_COLORS` is a `const` array in `src/shared/ui/css-color.ts`, `CssColor` derives from it, and `styles/_tokens.scss` is generated from it:

```scss
// Generated from src/shared/ui/css-color.ts
@mixin colors(
  $background,
  $foreground,
  $border-hairline,
  $border-hairline-strong
) {
  --color-background: #{$background};
  …
}
```

Call sites are untouched — still positional, in array order — and the compiled CSS is identical.

**On your `--check` point, you are right about the intent and the mechanism does not hold.** Two things, one of which bit me while implementing it:

1. `report_autofix` bails when the tree was *already* dirty (`[ -z "$STATUS_BEFORE" ] || return 0`) and never touches the exit code. My first attempt did the before/after comparison in `vet.sh` instead — which is also wrong, because `git status --porcelain` reports *which* files are modified, not their content: on a branch where the partial is already modified, the before and after strings are identical and drift sails through. Verified.
2. More decisive: a write **inside** the fan-out races `lint:css` and `format:check`, which glob those same two `.scss` files. `writeFileSync` truncates before it writes, so a concurrent reader can catch one empty — the same class of intermittent failure the build-runs-alone comment exists for.

So the generator signals it itself: **it writes the partials and exits non-zero exactly when it had to.** One run repairs the drift and reports it, `git diff` is the report, a re-run is green — your two-for-one, with no dependence on tree state. `--check` is gone; there is one mode. `vet.sh` runs it between the build and the fan-out, and a run that finds nothing stale writes nothing at all, so the common case stays read-only.

Both drift directions verified, plus the third net:

| probe | result |
| --- | --- |
| hand-edit `_tokens.scss` | rewritten, exit 1 |
| add a token to `CSS_COLORS`, don't regenerate | rewritten, exit 1 |
| re-run either | exit 0 |
| add a token *and* regenerate | `Error: Missing argument $accent` — the Sass build fails at every `@include tokens.colors(…)` that does not pass it |

That last one is why adding a token cannot half-land: the mixin gains a required parameter rather than defaulting.

`CLAUDE.md` and `.claude/rules/styling.md` record the new shape, including why this is the one step in vet allowed to rewrite the tree — a generated partial has exactly one correct content, so there is no judgment in the fix the way there is in `eslint --fix`.

---

vzakharov:
### Attestation

Verified `57639ef` (`claude/migrate-mantine-tailwind-w0kion`) against `main` @ `7e8b546` "feat: an opaque chart, a rasterized OG card, and richer Playgram cuts (pr #20)".

- `./scripts/vet.sh` — **green** (build, styles:codegen, typecheck, eslint, prettier, stylelint, steiger, type-overlap, content:og --check, node --test)
- CI on this PR — **nothing runs on a PR in this repo** (`.github/workflows/deploy.yml` is `push`-only)
- CI-only buckets — **`/test-on-gh` is not present in this project**, so none was dispatched

Beyond vet, the rendered-markdown regression this branch introduced was verified numerically rather than by eye: both revisions were rendered under headless Chromium and their computed `font-size`, `line-height`, `font-weight`, `font-style`, margins, padding, colour, `text-decoration` and border values compared element by element across `p`, `a`, `strong`, `em`, `h1`–`h4`, `ul`, `ol`, `li`, `blockquote`, `hr`, inline and block `code`, `pre`, `table`, `th`, `td` and the image/video/mermaid/table-scroll wrappers, plus each image's rendered box against its intrinsic ratio. Every value now matches production to within the four-decimal rounding stylelint enforces; the one residual difference is `color` on an `hr` whose border colour is set explicitly, so it is inert.

`57639ef` nests `prose.scss` under its scope selector. It was held to the same standard from the other direction: both revisions of the sheet were compiled with `sass` and their rule sets compared, and all 58 rules are identical apart from the deliberate Mermaid print fix the commit describes.

`main` @ `7e8b546` was merged into the branch and the merge check found no further movement; the PR is `MERGEABLE` / `CLEAN`.

---

PR #20, opened 2026-09-02
feat: an opaque, rasterized Open Graph card for the Playgram cuts

vzakharov:
## Summary

- **The units-of-work chart gets an opaque plate.** The chart is an ``, and an SVG inside `` resolves `prefers-color-scheme` against the OS rather than the page — so the site in dark mode on an OS set to light was already rendering `#0b0b0b` text on `#000000`, and a paste into a dark document did the same. A full-bleed `<rect>` painted first, filled from the same media query as the palette, carries `globals.css`'s own `--background` values: invisible when the OS and the surface agree, a deliberate figure when they don't.
- **The chart becomes the Open Graph card for all three cuts.** No major consumer renders an SVG `og:image` — X, Facebook, LinkedIn, Slack and iMessage all drop it and fall back to nothing — and a static export has no request-time renderer, so `pnpm content:og` rasterizes it into a committed 2400×1260 PNG. `--check` recomputes the source's hash against `og-renders.json` and runs in `vet.sh`, so an edited chart cannot ship behind a stale card. The card's dimensions are read from the PNG and published, replacing the assumption that a custom image's size is unknowable.
- **The mini and nano cuts get non-prose assets.** Both read as a wall of text, and the chart — the evidence for the claim each of them leads with — appeared only in the full document. mini gains four assets, nano three; nothing is newly authored, and the copied mermaid fence is byte-identical so both documents share one committed render pair.
- **`micro` is renamed `nano`**, so the two cuts' leading letters differ. `VARIANTS` is the source of truth, so the routes, `generateStaticParams`, `siblingVariants` and the sitemap all follow. `/case-studies/playgram-bubble-to-nextjs-part-1/micro` now 404s — a static export on GitHub Pages cannot redirect, and the study is days old with no known external links.
- **Three helpers gain a second caller and move to one home each**: the PNG/SVG header reader to `shared/content/image-dimensions.ts`, `mermaidHash` to `contentHash` in `shared/content/content-hash.ts`, and `findChromium` to `scripts/lib/chromium.ts`. The content-tree walk both render scripts duplicated becomes `contentFiles` in `scripts/lib/content-tree.ts`.

## QA Checklist

- [ ] `chart-mismatch` — open a cut with the **site theme and the OS colour scheme disagreeing** (site dark / OS light, then site light / OS dark). The chart must sit on its own plate and stay legible; it must never render dark text on a dark surface.
[...]

PR #21, opened 2026-09-02
fix: reach the GitHub API when the egress proxy refuses it

vzakharov:
## Summary

- **Syncs the agent boilerplate forward** from `eab4dc7` to `7f3ca76` (`vzakharov/agent-project-boilerplate`). Three source commits touched adopted paths and all three landed — one as a near-verbatim port, two translated against a `CLAUDE.md` that had diverged. No new skills appeared upstream, so there was nothing to offer under the "new skills get offered, not taken" step.
- **`export-github-item.py` and `pr-body.py` now survive the egress proxy.** Both reached the GitHub API through a bare `urlopen`, which honors `HTTPS_PROXY` — and a Claude Code remote session's proxy refuses `api.github.com` with its own 403, so the first API call killed either script outright. The proxy-then-direct ladder that already existed for attachment downloads moves into `scripts/lib/github.py` and serves all three call sites through one `fetch`. The proxy rung stays first, so a session where the proxy is the only route out is unaffected.
- **`CLAUDE.md` forbids resolving GitHub comment threads.** Thread resolution is the reviewer's tracking mechanism — they read down the replies and resolve what satisfies them, leaving the rest as the list of what still needs attention. The rule states that it overrides the contrary harness instruction, which is load-bearing rather than boilerplate: that instruction appears in every session that drives a PR.
- **`CLAUDE.md` states where an approved lint suppression goes**, as a sibling bullet in "Key principles". This repo had the point-of-use instruction tacked onto the gate bullet with no rationale, and the reasoning only in `.claude/rules/eslint.md` — which loads solely when `eslint.config.ts` or `eslint/**` is touched, so an agent reaching for a `# type: ignore` under `scripts/` never saw it. The rule file keeps what is genuinely ESLint-scoped and cites the principle instead of restating it.

### Triage

| --- | --- | --- |
| `3f992e2` proxy ladder | **take** | Both scripts and `lib/github.py` are adopted verbatim here and were byte-identical to the source's parent bar two docstring wordings this repo had tightened; those local wordings are preserved. |
| `03401ac` never-resolve | **take** | This repo's "GitHub comments" section was verbatim-identical and had no such rule. |
| `7f3ca76` suppression placement | **translate** | This repo had independently reached the point-of-use half (PR #16) but scoped its rationale to an ESLint-only rule file. Took the missing rationale and corollaries at stack-neutral altitude; de-duplicated the rule file against them. |

`ADOPTING.md` was the one unfamiliar path the unfiltered log surfaced. It predates the last sync and is the source's guide for a repo taking the boilerplate on — a thing this repo did once and will not do again — so it joins `declined` to keep the next sync quiet.

## QA Checklist

- [ ] `export-issue` — with `HTTPS_PROXY` set as the session sets it, run `python3 scripts/export-github-item.py 17` and confirm it writes `docs/issue/17/issue.md` instead of dying on a 403. Before this change the first API call was fatal.
- [ ] `export-attachments` — run the exporter against an issue carrying an image attachment and confirm the file lands under `attachments/` with a real extension.
- [ ] `pr-body-roundtrip` — `python3 scripts/pr-body.py pull <n>`, edit a word, `push <n>`, and confirm the PR body on GitHub reflects the edit.
- [ ] `failure-message` — point `fetch` at a URL both rungs refuse (e.g. a nonexistent repo) and confirm the error lists each rung's status *and* body, so the proxy's 403 names itself as the refuser.
- [ ] `token-not-leaked` — force an attachment download to fail and confirm the message carries status lines only, with no S3 body echoing the bearer token back.
- [ ] `suppression-guidance` — read the new "Key principles" bullet against `.claude/rules/eslint.md` and confirm the two layer rather than repeat: the principle and its rationale in `CLAUDE.md`, the three ESLint spellings in the rule file.
- [ ] `watermark` — confirm `.claude/skills/sync-agent-boilerplate/source.json` parses, names `7f3ca76`, and that `bash scripts/check-skill-catalog.sh` passes.

| --- | --- | --- | --- |
| `export-issue` | integration | ❌ | Run the real script against a fixture repo with `HTTPS_PROXY` set; assert exit 0 and the written path. No Python tests exist yet — `pnpm test` is Node-only. |
| `export-attachments` | integration | ❌ | Same harness, an issue with an attachment; assert the file exists and its extension was sniffed. |
| `pr-body-roundtrip` | integration | ❌ | Needs a live PR to PATCH; cheapest as a scripted round-trip against a scratch PR. |
| `failure-message` | unit | ❌ | Stub both openers to raise `HTTPError`; assert `format_route_statuses_and_bodies` renders both rungs. |
| `token-not-leaked` | unit | ❌ | Assert `format_route_statuses` output contains no body text for a failure carrying one — the security-relevant row. |
| `suppression-guidance` | manual-only | — | Whether two prose layers read as complementary is a judgment call. |
| `watermark` | unit | ❌ | `json.load` plus a SHA-shape assertion; `check-skill-catalog.sh` already covers the reference closure half. |

Both scripts were smoke-tested end-to-end in this session under the session's own proxy: the exporter wrote `docs/issue/17/issue.md`, and `lib.github.fetch` reached `api.github.com` successfully. The rows above are the standing regression gaps, not unverified behavior.

https://claude.ai/code/session_017JKxSPaAN9BdtmW1UYB9Xu


vzakharov:
Proposed squash title/body:

```
fix: reach the GitHub API when the egress proxy refuses it (pr #21)
```

```
`export-github-item.py` and `pr-body.py` reached the GitHub API through
a bare `urlopen`, which honors `HTTPS_PROXY` — and a remote session's
egress proxy refuses `api.github.com` with its own 403, so the first API
call killed either script outright. Both were dead on arrival in the
environment they were written for.

The proxy-then-direct ladder that already existed for attachment
downloads moves into `scripts/lib/github.py` and serves all three call
sites through one `fetch`, which takes a request factory because
`ProxyHandler.proxy_open` rewrites a request's host in place. The proxy
rung stays first, so an environment where the proxy is the only route
out is unaffected. Failures render two ways on purpose: statuses alone
for the attachment path, where S3 quotes the bearer token back in full,
and statuses with bodies for the API paths, where the proxy's 403 names
itself as the refuser.

Two rules join CLAUDE.md alongside it. Resolving a review thread is
forbidden outright: resolution is the reviewer's tracking mechanism, and
a thread the agent tidies away drops off their list whether or not they
read it — stated as overriding the contrary harness instruction, which
every PR-driving session sees. And an approved lint suppression belongs
at the point of use rather than the linter's config, because a config
that carries domain concerns stops being portable and a central entry is
invisible from the code it exempts. This repo had the instruction
without the reasoning, and only inside `.claude/rules/eslint.md`, which
loads solely for `eslint.config.ts` and `eslint/**`; the principle moves
up to where every linter in the repo is in scope, and the rule file
keeps the ESLint spellings.

All three changes arrived from vzakharov/agent-project-boilerplate,
whose watermark advances from eab4dc7 to 7f3ca76.

```

---

vzakharov:
### Attestation

Verified `f71fc9f` (`claude/sync-agent-infra-upstream-m7zlsv`) against `main` @ `7e8b546` "feat: an opaque chart, a rasterized OG card, and richer Playgram cuts (pr #20)".

- CI on this PR — **nothing runs on a PR in this repo** (`deploy.yml` triggers on push to `main` only)
- CI-only buckets — **not dispatched: `/test-on-gh` is not part of this project's skill set**

Two things the reviewer should weigh it against:

**The vet suite does not reach this PR's main subject.** Every check in `vet.sh` is TypeScript/Node — build, typecheck, eslint, prettier, steiger, type-overlap, `content:og --check`, and a `pnpm test` that globs `**/*.test.ts`. The bulk of this diff is Python (`scripts/lib/github.py`, `scripts/export-github-item.py`, `scripts/pr-body.py`), which no check in the repo executes. A green vet here means the merge with `main` is sound and nothing in the TypeScript tree regressed — it says nothing about the scripts. Those were instead **smoke-tested end-to-end in-session under the session's own egress proxy**: the exporter wrote `docs/issue/17/issue.md`, and `lib.github.fetch` reached `api.github.com` successfully through the ladder. The PR body's coverage table records the standing regression gap.

**Vet ran on `9950750`, the merge commit.** Everything after it (`f4a3078`, `76d9d8d`, `f71fc9f`) only writes and then deletes `docs/remove-before-merging/squash-message.md`, which no check reads and which does not reach `main`.

The base merge was clean and the incoming diff (content pipeline, OG rasterization, `src/shared/content/**`) shares no path or symbol with this branch. It did add a seventh concurrent check to `vet.sh`, so the suite was re-run against the merged tree rather than trusting the pre-merge pass.

---

PR #26, opened 2026-09-05
docs: a LinkedIn content plan mined from the Playgram case study

vzakharov:
## Summary

- **`writing/linkedin/plan.md`** — the standing backlog: an eighteen-post inventory, each row anchored to one concrete piece of work; the suggested opening run; the PDF attachment the announcement carries; an **Unsettled** section for the four rows that hold an objection rather than a draft; and a "not yet" reserve so the backlog outlives the current material.
- **`.claude/rules/writing.md`** — the per-draft contract, path-scoped to `writing/**`: the 1,000-character ceiling and the ~200-character fold, the five shapes, the frontmatter every draft carries, the voice conventions, and the copy-exactness of a post's fenced text.
- **`writing/linkedin/drafts/case-study-announcement.md`** — the first draft. Its text is the operator's own, filled at one blank (362 lint rules) and otherwise verbatim; it is the specimen the voice rules are written against.
- **`writing/notes/the-five-percent.md`** — evidence gathered for a post that isn't draftable yet: cases where a human review caught what an agent could not. Keyed on the abstracted learning rather than the review, each carrying `(×n)` bumps and sorted by it, with the squeeze order written in for when it outgrows itself. Adding to it after a review is mandatory until the post it feeds runs.
- **CLAUDE.md** gains the `writing/` layout row, a `content:` commit prefix — deploy-free by definition, so `.github/workflows/deploy.yml` is untouched — and, in § "GitHub comments", the rule that a review session owes an entry to the five-percent file. The `writing/` tree sits at the repo root rather than in `public/` (served — an unposted draft crawlable before it runs is the wrong order of operations) or `docs/` (whose subdirectories are working artifacts `/finalize` sweeps).

## What the first review changed

16 comments, addressed in `d14fe1d`; every thread has a reply.

- **The announcement is the operator's rewrite, verbatim.** The agent-picked skim baits — a lint rule whose failure mode is an auth hole, two type shapes that both compiled — are replaced by the three the operator chose, which are legible to someone who wasn't there: a number to react to, a change in how the work felt, an admission. Two apparent dropped words are flagged on the thread rather than silently kept.
- **The rule file absorbed everything that holds for every draft** — the Form section entire, the five shapes, the frontmatter contract — leaving `plan.md` the backlog and the strategy.
- **The shapes spell FRAME.** `Receipt` became `Erratum`; the order follows the mnemonic. The mix is unchanged at 3/2/1/2/2.
- **Three voice rules were extracted from the rewrite**: paragraphs break at the turns, sentences don't get tightened, and the hook is the operator's call — an agent reaches for the technically impressive detail where a reader wants the legible one.
- **Inventory**: the `safe-action` row became the `safeAction` machinery itself; the skills row became `agent-project-boilerplate`; the stale-plan row widened to how much documentation is the right amount.
- **The parallelism numbers are withdrawn** pending #28 — both windows behind 6.2 → 8.2 are mis-drawn.

No longer blocked: #25 landed the PDF pipeline, so `public/case-studies/playgram.pdf` is on `main` and the `blocked-on` key is gone. `main` is merged in; the anticipated CLAUDE.md layout-table conflict was resolved by keeping both edits.

## What the second review changed

5 comments, addressed in `4a0affe`; every thread has a reply.

- **`content:` is a new commit prefix**, because none of the existing ones describes a change to the site's own written material — `docs:` is documentation, `feat:` is the code that serves it. `CLAUDE.md` now also records that the prefix list is local and extensible, and `/squash-message` reads the project's list instead of assuming the conventional-commit set.
- **`writing/notes/` exists**, and row 18 points at it instead of carrying one example inline. Its first file has specimens from every review of this branch, splitting two ways: those where the reasoning inside the frame was sound and the frame itself was the thing to move, and those where the output was well-formed, justified and efficient and every one of those properties is what made it wrong.
- **Row 8's line counts are now unverified**, not just their reading. 98,000 → 223,000 over the 81 days from `4.0.0` to handover is published rather than checked, and re-deriving it needs the Playgram repo. Not filed as an issue since only the chart was asked for — say the word and it joins #28.
- **"English only" is dropped.** Which language a post is written in is a per-post call, and the rule anticipated a question that had not come up.

## What the third review changed

5 comments, addressed in `2c1108c`; every thread has a reply.

- **`content:` does not publish**, and the dilemma the second pass reasoned through does not exist: the prefix is *specifically* for written material the site does not yet serve, and a piece going live arrives as the page that serves it — `.md` and `.pdf` variants included — which is a `feat:`. `.github/workflows/deploy.yml` is byte-identical to `main` again, and this PR spends no deploy.
- **An entry in `writing/notes/the-five-percent.md` after a review is mandatory**, stated in `CLAUDE.md` § "GitHub comments" where review handling already lives. Both the rule and the file name their end: they retire when backlog row 18 is posted, which is what keeps a mandatory notes file from becoming a standing ritual. `.claude/rules/writing.md` says the same for `notes/` generally.
- **The squash proposal is the proposal and nothing else** — the prefix paragraph and the two notes beside it are gone: a minor ride-along, and implementation detail either way.
- **This review's own two specimens are recorded**, one per family: a word the agent had coined the day before and then reasoned about as if its meaning were given, and a commit body made worse by documenting everything true about the change.
- **The five-percent file is keyed on learnings, not reviews** (`24f2d1e`, on follow-up). Six learnings over nine bumps, `(×n)` in each heading and the list sorted by it, so the post is drafted from the top rather than from the most vivid specimen. Counts are maintained by hand — deciding that two reviews bumped into the same thing is the judgement the file is made of. Past ~400 lines the file squeezes rather than grows: tighten, then cut archaeology to the shortest account that still shows what the agent could not see, then drop learnings from the bottom.

Prose and one CLAUDE.md section — nothing here touches `src/`, `scripts/` or `public/`, and the one workflow edit the branch had is reverted.

## QA Checklist

- [ ] `post-text` — open `writing/linkedin/drafts/case-study-announcement.md` and read the fenced block as the post that will actually go out. It should be your text, unchanged apart from `362` and the two flagged word fixes.
- [ ] `flagged-edits` — confirm `abstractions and 11.6-megabyte` and `how much faster switching` are the intended readings, or say which to revert.
- [ ] `ceiling` — the fenced text is 573 characters against the 1,000 ceiling, in three paragraphs; the first is 186, so the whole opening clears the ~200 fold.
- [ ] `frame` — the five shapes spell FRAME in `.claude/rules/writing.md`, and every `Shape` cell in the inventory uses one of the five (no `Receipt` left).
- [ ] `split` — `plan.md` no longer carries Form, the shapes or the frontmatter table, and the rule file carries each exactly once.
- [ ] `unsettled` — the four ⚠ rows (3, 8, 11, 18) each have a matching entry under "Unsettled", and row 11 no longer asserts 6.2 → 8.2.
- [ ] `inventory` — skim the rewritten rows (2, 3, 10, 11, 15, 18) for any claim that overstates what the case study supports.
- [ ] `specimens` — read `writing/notes/the-five-percent.md` as a record of what you actually did in these three reviews, and correct anything it credits you with that isn't yours.
- [ ] `bump-counts` — each `(×n)` matches the bumps listed beneath it, the sections descend by count, and the ~400-line squeeze threshold is the one you'd pick.
- [ ] `content-gate` — `git diff main -- .github/workflows/deploy.yml` is empty, and the squash subject `content:` therefore lands without a deploy.
- [ ] `five-percent-rule` — read the two new paragraphs in CLAUDE.md § "GitHub comments" and confirm the obligation is what you meant, and that its expiry reads as binding rather than as a hedge.
- [ ] `copy-paste` — select the fenced block and paste it somewhere plain; confirm three paragraphs arrive with no mid-sentence line breaks, `--` intact and the 🙈 surviving.
- [ ] `pdf` — open `public/case-studies/playgram.pdf` and look at it; a print render fails by being ugly, not by failing a gate.
- [ ] `layout-row` — the `writing/` row renders in CLAUDE.md's repository-layout table beside main's rewritten `public/` row, and its `@.claude/rules/writing.md` pointer resolves.

| `post-text` | manual-only | — | Voice judgment against a specimen the operator wrote; nothing to assert |
| `flagged-edits` | manual-only | — | Only the author can say whether a dropped word was intended |
| `ceiling` | unit | — | A character count over the fenced block is assertable, but one draft does not justify a checker |
| `frame` | unit | — | Cross-checking shape cells against the rule file is scriptable; two files do not justify it yet |
| `split` | manual-only | — | Whether a convention sits in the right file is a judgment about scope |
| `unsettled` | unit | — | Same shape as `frame`: assertable, not yet worth a script |
| `inventory` | manual-only | — | Editorial judgment on claim strength and fit |
| `specimens` | manual-only | — | A record of the reviewer's own reasoning; only they can verify it |
| `bump-counts` | unit | — | Count-vs-children and sort order are assertable, but the file is hand-kept on purpose and retires with the post |
| `content-gate` | unit | — | A diff against `main` is assertable, but a one-line workflow revert does not justify a check |
| `five-percent-rule` | manual-only | — | Whether a written obligation says what the operator meant is theirs to judge |
| `copy-paste` | manual-only | — | Behavior of the clipboard and the LinkedIn composer, outside the repo |
| `pdf` | manual-only | — | Visual judgment on a print rendering |
| `layout-row` | manual-only | — | Markdown table rendering and a doc-internal path reference |

https://claude.ai/code/session_011DB8h6Gknzp5WbMvUCk6KC


vzakharov:
Proposed squash title/body:

```
content: a LinkedIn backlog mined from the Playgram case study (pr #26)
```

```
The Playgram case study is a 40-minute read, and most of what is in it
will only ever reach anyone as something shorter. This lays down the
backlog and the form up front, so posts get drafted against a standard
rather than invented one at a time.

`writing/linkedin/plan.md` is the standing backlog: eighteen posts, each
anchored to a fact specific enough that it could not have been written
by someone who wasn't there, the opening run, a reserve list for when
the current material thins, and an "Unsettled" section for the three
rows that carry an objection rather than a draft.

`.claude/rules/writing.md` is the per-draft contract, scoped to
`writing/**`: the 1,000-character ceiling and the fold, the five shapes
a post can take, the frontmatter every draft carries, and the voice
conventions -- `--` over an em dash, self-deprecating emoji only, and
the copy-exactness of a post's fenced text, which is one unwrapped line
per paragraph because a hard wrap in the source becomes a line break in
the feed.

The first draft is the announcement, carrying the case study as a PDF
attachment rather than a link. Its text is the operator's own, filled
at one blank, and it is the specimen the voice rules are written
against.

`writing/notes/` holds evidence gathered for a post that isn't
draftable yet. Its first file collects cases where a human review
caught what an agent could not -- the claim one backlog row makes and
cannot yet support -- keyed on the abstracted learning rather than the
review and sorted by how many reviews bumped into each, so the post is
written from the top of that list.

`writing/` sits at the repo root deliberately: `public/` is served, and
an unposted draft crawlable before it runs is the wrong order of
operations, while `docs/` is where working artifacts go to be swept.

Claude-Session: https://claude.ai/code/session_011DB8h6Gknzp5WbMvUCk6KC
```

---

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,44 @@
+Proposed squash title/body:
+
+```
+docs: a LinkedIn backlog mined from the Playgram case study (pr #26)
```
vzakharov: that's not exactly docs, let's approach per substance

docs/remove-before-merging/squash-message.md
```diff
@@ -0,0 +1,44 @@
...
+and holds what applies to every draft -- today the two settled
+conventions, since the anti-slop rules are meant to be written against
+marked-up drafts rather than anticipated.
+
+`writing/` sits at the repo root deliberately: `public/` is served, and
+an unposted draft crawlable before it runs is the wrong order of
+operations, while `docs/` is where working artifacts go to be swept. The
+announcement attaches the case study as a PDF, which
+`claude/content-url-suffixes-j2qjau` owns rather than this branch, so
+the draft is held by a `blocked-on` key until that lands.
```
vzakharov: no need for any of this; by the time we'll be merging it'll be on main, so this is just for intra-pr understanding

writing/linkedin/drafts/case-study-announcement.md
```diff
@@ -0,0 +1,40 @@
+---
+source: whole document
+shape: announcement
+status: draft
+blocked-on: claude/content-url-suffixes-j2qjau
```
vzakharov: same

writing/linkedin/drafts/case-study-announcement.md
```diff
@@ -0,0 +1,40 @@
...
+
+Everything but the three phrases below is the operator's own. This is the voice
+specimen the writing rules calibrate against, so it does not get tightened,
+rebalanced or improved.
+
+603 characters, against the 1,000 ceiling; the fold at ~200 falls on the
+sentence break, so the JSON lands above it.
+
+```text
+So from March to August this year I was doing quite a fun thing -- well, fun if you're the weird kind who loves dabbling with parsings, abstractions and 11.6-megabyte JSONs that live on a single line. Once done, I thought, why not write a case study about it. I only went halfway and it's already a 40-minute read 🙈 Not really thinking anyone would seriously read it all -- but maybe skim through? There's some interesting stuff on a lint rule whose failure mode was an authentication hole, two type shapes that both compiled while logging zeroes for months, and managing a team of 15+ agents at a time.
```
vzakharov: Rewritten, take verbatim except for <placeholders> blanks and extract to rules where extractable (don't overengineer though).

```
So, from March to August this year I was doing quite a fun thing -- well, fun if you're the weird kind who loves dabbling with parsings, abstractions a 11.6-megabyte single-line JSONs.

Once done, I thought, why not write a case study about it. I only went halfway and it's already a 40-minute read 🙈

Not really thinking anyone would seriously read it all -- but maybe skim through? There's some interesting stuff on why I decided to turn on <count> linter rules, how faster switching from CLI to Web agents made me, and some embarrassing mistakes I made along the way.
```

So basically less fancy writing, plus the stuff you chose didn't seem that fun to me. (A meta side note it's fun in itself to see how an agent picks "what's interesting" differently from a human.)

writing/linkedin/drafts/case-study-announcement.md
```diff
@@ -0,0 +1,40 @@
...
+
+Attaches `playgram.pdf`. No link in the body — the file is right there. No
+sign-off line.
+
+Everything but the three phrases below is the operator's own. This is the voice
+specimen the writing rules calibrate against, so it does not get tightened,
+rebalanced or improved.
+
+603 characters, against the 1,000 ceiling; the fold at ~200 falls on the
+sentence break, so the JSON lands above it.
```
vzakharov: This mostly looks like PR narration; what we need is to say what we decided to write about and why.

writing/linkedin/drafts/case-study-announcement.md
```diff
@@ -0,0 +1,40 @@
...
+  failure mode is a live authentication hole, because Next turns every export
+  from a `'use server'` file into an endpoint the browser can call.
+- **second skim bait** — `tokenCounts: { input, output }` sitting beside DB
+  columns named `inputTokens`/`outputTokens`. Both type-checked; every usage log
+  recorded zero, found by accident months later.
+
+An announcement calibrates the personal register only, and says nothing about
+the teaching register the other eighteen posts live in. Post #3, the
+type-overlap gate, is the natural second calibration draft once the first round
+of rules exists.
```
vzakharov: don't think we'll need this section at all; such overspecifying tends to accumulate over time, leading to posts leaning into one direction. If anything, it belongs to the plan (which will be removed) or even the session chat, where we discuss what to write about and why.

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+piece of work — mostly from the Playgram case study, later from whatever else
+ships. Every post leaves a practitioner with something they can use on Monday.
+Nothing is a teaser for a longer thing; a post that only works as an
+advertisement is not a post.
+
+Two effects are wanted and **neither is ever named in the text**: that a reader
+who needs this kind of work done starts wondering who does it, and that the name
+becomes one people recognise on the subject of running coding agents at scale.
+Both come from the writing being useful and specific. No CTA, no "what's your
+experience?", no "DM me". The dots stay for the reader to connect.
```
vzakharov, line 13: Add that although it's likely that this decreases engagement, this is a long-term "branding" bet to sound authentic and not needy -- unlike the majority of such posts.

writing/linkedin/plan.md
vzakharov, line 1: a lot of this (e.g. the Form section in its entirety) belongs  to the claude rule, not the plan

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+  it now. Ends on the mechanism, never on a moral.
+- **Measurement** — a number nobody else has, and what it does and doesn't mean.
+  Must include the deflating half; a measurement that only flatters is an ad.
+- **Reversal** — a received best practice that is wrong in a named context, with
+  the reason it's usually right left standing.
+- **Receipt** — an unflattering fact about my own work, scored. It does the most
+  for credibility, and must never be false modesty about something that actually
+  went well.
+- **Artifact** — a small piece of real code, config or output, shown and then
+  explained. The reader should be able to steal it.
```
vzakharov: As much as I hate this (lol), but the initials almost add upp to F.R.A.M.E., except that we have to R's and no E's. See if we can make it fit entirely.

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 1   | Splitting by shape, not by size                        | An 11.6 MB minified JSON export cut into 3,487 files an agent can _navigate_ instead of grep. Names derived from content, never position, so inserting a step doesn't rename its neighbours; long strings hoisted into `.txt` siblings so a prompt diffs as prose instead of one enormous escaped line. The point is the weekly re-export staying legible. | Artifact    |
+| 2   | The lint rule that closes an auth hole                 | Next turns every export from a `'use server'` file into an endpoint the browser can call — no route file, no ceremony. `safe-action-required` forces every one through a wrapper that establishes who is asking before the body runs. An agent will cheerfully ignore a paragraph in CLAUDE.md and will never once ship a lint error.                      | Footgun     |
```
vzakharov: this one's boring; i'd rather write about the safeAction machinery per se: making the back-to-front API work naturally, where errors thrown are converted to object keys, in an elegant way, where everything is logged at the right places (making the /log-review skill possible), etc. etc.

(you can download that repo for a better insight via the gh shim btw, just in case you haven't figured that out already.)

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 7   | Parallel agents meet sequential migration numbers      | Migration `0099` generated on a branch that hadn't merged `0098`. Forked chain, dropped enum value, nothing failed at deploy for two whole migrations. Now a script walks the snapshots and refuses a broken chain.                                                                                                                                        | Footgun     |
+| 8   | Rigid boundaries thicken the shared layer on their own | Between first production build and handover `src/` went 98,000 → 223,000 lines, and the layers that grew fastest _in relative terms_ were the bottom ones: `shared` and `entities` nearly tripled while the app-specific top layer didn't quite double. Nobody planned that.                                                                               | Measurement |
```
vzakharov: I'm still on the fence about that one: first of all, I'm not sure that "if your shared layer grows, that's good" is correct. But I need to think some more about it when the time comes.

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 9   | Five agents told me it would be fine                   | Four models researching every decision independently, names stripped so the synthesis couldn't be biased, a fifth judging. The lone dissenter won both contested points. And the honest half: a good chunk of it was spreading the blame. Verdict 6.5/10.                                                                                                  | Receipt     |
+| 10  | The plan listed a directory that never existed         | The detailed migration plan's own "current state" inventory named an `entities/message` slice that was never in the repo, not once, in its entire history. After a while you're not reading the plan to find out what to do — you're reading it to find out how out of date it is. Verdict 6/10.                                                           | Receipt     |
```
vzakharov: I'd widen it to a general look on how much docs is good

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 11  | What parallelism actually measures as                  | Across the one-day switch into the cloud: median unit of work 375 → 384 changed lines, units per day 6.2 → 8.2. Same-sized pieces, about a third more of them at a time. That's the number, and it is smaller than the marketing.                                                                                                                          | Measurement |
```
vzakharov: I actually meant to comment on/oppose this measurement, see here:

Image

The "after transition to web" should be measured according to the two green lines: in-between them there was a considerable number of days off, which shouldn't affect the measurement; and after 4.1.0 shift it was a different context altogether (before 4.1.0 = ship the Bubble-parity app, after that = bug fixes, and later on, new features). For completeness, the red arrow, too, should start not at the very beginning -- when it was a docs-only phase.

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 14  | The merge conflicts I was wrong about                  | The loudest fear about parallel agents, and after a thousand-plus merged PRs, zero problems — once one skill encoded the footguns. Agents resolve conflicts semantically, not just to a state where the file no longer has markers in it.                                                                                                                  | Receipt     |
+| 15  | Thirty-three files that describe how you work          | Skills compose into a call graph because each is a file another can point at — `/implement` loads two by reference and hands to a third, which loads three more. And the half that's less fun: it's a second codebase, it drifts exactly like the first, and nothing lints it.                                                                             | Receipt     |
```
vzakharov: ykw, I'd rather talk about https://github.com/vzakharov/agent-project-boilerplate, which was born out of the Plagram project and is now an open-source "agent starter pack" for anything I code (and others are welcome to use it too -- at Playgram management's courtesy, because technically it too is a Playgram asset)

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 17  | Fail slow, when the reader is an agent                 | Every check runs to completion even after one has failed, and the summary names all of them. Fail fast and the agent fixes one thing, pushes, and waits four minutes to be told about the next. Received wisdom, inverted for a reason.                                                                                                                    | Reversal    |
+| 18  | Reviewing the outcome, not the process                 | From three local agents whose _thinking_ I read, to twenty in the cloud reviewed as pull requests. For 95% of cases the agent knows better; the remaining 5% is the entire argument for a knowledgeable human still being in the loop.                                                                                                                     | Receipt     |
```
vzakharov: That'll frankly be the hardest to write about because even know I can't quite pin down what the "5%" are. Just saying. Maybe this code review session is a meta-example through which we can see where the human sees something the agent doesn't, and go from there.

writing/linkedin/plan.md
```diff
@@ -0,0 +1,154 @@
...
+| 2   | The lint rule that closes an auth hole                 | Next turns every export from a `'use server'` file into an endpoint the browser can call — no route file, no ceremony. `safe-action-required` forces every one through a wrapper that establishes who is asking before the body runs. An agent will cheerfully ignore a paragraph in CLAUDE.md and will never once ship a lint error.                      | Footgun     |
+| 3   | Two types, one lie, months of zero analytics           | `tokenCounts: { input, output }` beside DB columns named `inputTokens`/`outputTokens`. Both compiled. Every usage log recorded zero, found by accident months later. Then the second return nobody expects: about a third of reported overlaps are a _key that's lying_, not a missing base.                                                               | Footgun     |
```
vzakharov: This sounds hooky, but, just a hunch, maybe too hooky. I believe there's more substance to the type-overlap shenanigans than just catching one object type mismatch. Although yeah, it might be a prime illustration. Smth to still think about.

docs/remove-before-merging/squash-message.md
vzakharov: Switched to `feat:`. This documents nothing -- it's the repo starting to carry editorial content and the rules for producing it, and `feat:` is the closest of the allowed prefixes by substance. The trade is stated in the file: `feat:` trips the deploy gate for a change that alters nothing in the built site, so it costs one no-op Pages republish. Say the word and it goes back to `chore:`, which is deploy-free but undersells it.

---

docs/remove-before-merging/squash-message.md
vzakharov: Cut -- and moot anyway now that #25 landed. The squash body no longer mentions the sibling branch or the `blocked-on` key at all.

---

writing/linkedin/drafts/case-study-announcement.md
vzakharov: Removed. `public/case-studies/playgram.pdf` is on `main`, so nothing is holding the draft.

---

writing/linkedin/drafts/case-study-announcement.md
vzakharov: Taken verbatim, with `<count>` filled as 362. Two edits I made that I want you to veto if they're wrong:

- `abstractions a 11.6-megabyte` -> `abstractions and 11.6-megabyte`
- `how faster switching from CLI to Web agents made me` -> `how much faster ...`

Both read as dropped words rather than voice, but they're yours, so they're flagged rather than quietly fixed. 573 characters; the first paragraph is 186, so the whole opening clears the ~200 fold.

On extraction -- three rules came out of the delta, all in `.claude/rules/writing.md`: paragraphs break at the turns (mine was one block), sentences don't get tightened (the loose version is the voice), and the hook is yours to pick, because an agent reaches for the technically impressive detail where a reader wants the legible one. That last one is your meta note written down as a rule.

---

writing/linkedin/drafts/case-study-announcement.md
vzakharov: Replaced with what the post is and why: an announcement does one thing, the three baits are chosen to be legible to someone who wasn't there, and the PDF is attached rather than linked because an attachment opens in the feed. No drafting narration left.

---

writing/linkedin/drafts/case-study-announcement.md
vzakharov: Gone entirely.

---

writing/linkedin/plan.md
vzakharov: Added as its own paragraph -- the CTA is standard advice precisely because it moves the metric, and it's being traded away on purpose: a long-term bet on how the name reads a year from now rather than on how a post performs this week.

---

writing/linkedin/plan.md
vzakharov: Moved. `.claude/rules/writing.md` now takes the Form section entire, the five shapes, and the per-draft frontmatter contract -- everything that holds for every draft. `plan.md` keeps the backlog, the strategy, sequencing, the attachment and what's unsettled, and points at the rule file rather than restating it.

---

writing/linkedin/plan.md
vzakharov: It fits. `Receipt` -> **Erratum** ("a correction to my own record: something I got wrong or oversold, scored"), and the list is reordered **F**ootgun / **R**eversal / **A**rtifact / **M**easurement / **E**rratum. The mix comes out unchanged at 3/2/1/2/2 over ten.

---

writing/linkedin/plan.md
vzakharov: Rewritten to the machinery. Row 2 is now the single door every server action goes through: `safeAction(...)` establishes who is asking before the body runs and turns a thrown error into a key on a result object the frontend can read -- and because everything funnels through one place, logging lands where it belongs by construction, which is what makes a `/log-review` agent possible at all. The lint rule is demoted to a footnote. Shape moves Footgun -> Artifact.

I wrote it from the case study rather than pulling the Playgram repo. Worth reading the actual `safeAction` before this one gets drafted -- tell me the repo and I'll do that then.

---

writing/linkedin/plan.md
vzakharov: Left in the backlog, marked with a warning sign and written up under a new **Unsettled** section: "a shared layer that grows is a good sign" is asserted rather than established and may simply be wrong; the numbers are real, the reading of them is what needs another pass. Nothing gets drafted from it until that's settled.

---

writing/linkedin/plan.md
vzakharov: Widened. Row 10 is now "How much documentation is the right amount", with the never-existed `entities/message` slice as the illustration rather than the subject, and the general point underneath -- a doc is worth only as much as whatever updates it, so the right amount is however much will actually be maintained. Shape moves Receipt -> Reversal.

---

writing/linkedin/plan.md
vzakharov: Filed as #28 rather than fixed here, per your follow-up. It carries all three corrections: the after-window measured between the two green lines and ending at `4.1.0` (before it, shipping Bubble-parity; after it, bug fixes and later features -- different work), the days off excluded from the denominator, and the before-window starting after the docs-only phase rather than at day zero.

Row 11 now withdraws the 6.2 -> 8.2 figure explicitly and blocks on that issue, since the post's whole claim is that the honest number is smaller than the marketing -- which needs it honestly windowed first.

---

writing/linkedin/plan.md
vzakharov: Swapped. Row 15 is now the boilerplate: the skills, rules and vet suite with everything Playgram-specific stripped out, the repo every new project starts from including this site, open-sourced at Playgram management's courtesy since technically it's their asset. Kept "it's a second codebase, it drifts exactly like the first, and nothing lints it" as the deflating half. Shape moves Receipt -> Artifact.

---

writing/linkedin/plan.md
vzakharov: Kept, marked unsettled, and your point is what the entry says: naming the 5% needs specimens rather than argument, and this review is the first one -- you cut the narration, picked hooks the agent had passed over, and rejected a measurement it had taken as given. Noted as something to collect more of before drafting.

---

writing/linkedin/plan.md
vzakharov: Widened to the gate itself: 26 days of tightening a notch at a time, thirteen landings, ~1,300 type-level file-changes, 248 overlapping groups cleared, ~330 shared bases that didn't exist before -- plus the third of reported overlaps that turn out to be a key that's lying rather than a missing base. `tokenCounts` stays as the opener but the row is marked unsettled, since whether it leads or merely illustrates is exactly the open question.

---

docs/remove-before-merging/squash-message.md
vzakharov: I meant more like work: -- like we do with e.g. school: and travel:

writing/linkedin/drafts/case-study-announcement.md
vzakharov: yes, do fix the typos pls:)

writing/linkedin/plan.md
vzakharov: That's one example, but I think we can find more. Let's create a document in the codebase with your reflection and introspection as per the result of this code review as an illustration of where a human thinks sideways -- we'll use it  when we get to writing that piece of content.

writing/linkedin/plan.md
vzakharov: "the numbers are real" will need to be double-checked too, they look a bit too much to me (98k to 223k in two months)

.claude/rules/writing.md
```diff
@@ -1,19 +1,80 @@
...
+- **The first two lines carry it.** LinkedIn folds at roughly 200 characters, so
+  the opening states the concrete thing rather than promising it. No "I want to
+  share some thoughts on…", no question-as-opener.
+- **One idea per post.** A second good idea is a second post.
+- **A few short paragraphs, broken at the turns** — where the post changes what
+  it is doing, not every sentence. One block is a wall; one sentence per line is
+  the house style of the genre and reads as the genre rather than as a person.
+- **Numbers stay unrounded.** 3,487 files, 11.6 MB, 8,123 imports. The precision
+  is the evidence.
+- **English only.** The site is bilingual; this isn't.
```
vzakharov: I'd say it depends; sometimes I want to write in Russian; I'd say it doesn't need to be mentioned at all -- we'll see on a case by case basis

docs/remove-before-merging/squash-message.md
```diff
@@ -12,33 +12,38 @@ rather than invented one at a time.
...
-Claude-Session: https://claude.ai/code/session_01FL6zdfWFDTC2jvazgWQQQR
+Claude-Session: https://claude.ai/code/session_01QDxBYFvDDJXaKQmFaAdrLV
 ```
 
+**On the prefix.** `docs:` was wrong — this documents nothing; it is the
+repo starting to carry editorial content and the rules for producing it.
+`feat:` is the closest of the allowed prefixes by substance, and the
+trade is that it trips the deploy gate for a change that alters nothing
+in the built site: one no-op Pages republish. Say the word and it goes
+back to `chore:`, which is deploy-free but undersells it.
```
vzakharov: let's introduce "content:"

docs/remove-before-merging/squash-message.md
vzakharov: Understood — and this is the useful part of the note, not the prefix itself. I treated the list as the conventional-commits spec and reasoned carefully inside it (`feat:` vs `chore:` vs `docs:`, all three wrong in different ways), when the list lives in our own `CLAUDE.md` and adding a word to it costs one commit. `content:` is in, per the other thread. `CLAUDE.md` now also says the list is local and extensible, so the next change that fits nothing gets a new row proposed rather than filed under the nearest wrong one. 4a0affe

---

docs/remove-before-merging/squash-message.md
vzakharov: Done in 4a0affe: `content:` is in the `CLAUDE.md` prefix list — the site's own written material, where `docs:` is documentation and `feat:` is the code that serves it — and `.claude/skills/squash-message/SKILL.md` now reads the project’s list instead of assuming the conventional-commit set. This PR’s squash subject is `content:`.

**One call I made inside that, worth vetoing if you disagree: `content:` publishes.** It is in the deploy gate’s regex alongside `feat|fix`. The prefix covers both content under `public/` (which a visitor sees) and drafts under `writing/` (never built), and the subject line is all the gate reads, so one of the two has to be wrong. Publishing is the cheaper way to be wrong: a redundant no-op republish nobody notices, versus a live site quietly stale behind `main` with a hand-run as the only remedy — and hand-runs depend on remembering. The visible cost is that this PR, which changes nothing built, spends a deploy on merge. Say the word and the regex drops back to `feat|fix`.

---

writing/linkedin/plan.md
vzakharov: Agreed, and done in 4a0affe — this was the better move than recording the objection and waiting.

`writing/notes/the-five-percent.md` is the collection; `writing/notes/` is a new sibling of the channel directories, for evidence gathered against a claim a post makes before there is enough of it to draft from. Row 18's "Unsettled" entry now points there instead of carrying one example inline, and the standing instruction is that the entry goes in while it is still accurate rather than being reconstructed at drafting time.

The first entry is both of your passes over this branch, six specimens, and they already split two ways rather than one:

- **The frame was negotiable** (three): the prefix list, the published 6.2 → 8.2 and 98k → 223k numbers, and the "English only" rule. In each the reasoning *inside* the frame was sound and a second agent checking it would have passed — the failure was not noticing that the frame was ours to move.
- **The tell is the polish** (three): the paragraph explaining the paragraph, the tightened sentence, the hook. Here the output was well-formed, justified and efficient, and every one of those properties is what made it wrong.

The second family is the more interesting half for the post, because there is no version of "be more careful" that catches it — which may be the closest thing to an answer for what the 5% is. One sitting is not a pattern, so the file says so and holds it loosely.

Also written into the file, for honesty: the flagged-not-fixed words in your verbatim text are a rule doing its job, not judgement, so they shouldn't later get counted as a specimen.

---

writing/linkedin/plan.md
vzakharov: Fair, and flagged rather than defended — 4a0affe records in "Unsettled" that row 8 now carries two blocking objections, not one: the reading ("a shared layer that grows is a good sign") *and* the figures themselves, which are published rather than checked. Both need re-deriving from the repo before either is repeated. The table row says so too, so nobody drafts from it by skimming.

Two notes on the window, since "two months" may be understating it: the case study measures between `4.0.0` (21 May) and handover (10 Aug) — 81 days, not ~60 — and it says `src/` rather than the whole repo, so generated files, migrations and snapshots may or may not be inside that count depending on how it was taken. Neither makes 125,000 lines in 81 days obviously right; they just widen the range it has to be checked against.

I did not file this as an issue since you only asked for one on the chart, but it is the same kind of thing as #28 — a number that can only be re-derived from the Playgram repo, which this session cannot reach. Say the word and it goes in, either as a second issue or as a line on #28.

---

.claude/rules/writing.md
vzakharov: Dropped in 4a0affe.

Worth recording why it was there, because it is a pattern rather than a one-off: asked to write a conventions file, I wrote conventions — including one for a question that had never come up. The format wanted filling and a rule is cheap to write, which is exactly what makes a premature one expensive. Silence was available and I didn't see it. That went into `writing/notes/the-five-percent.md` as a specimen.

---

docs/remove-before-merging/squash-message.md
vzakharov: No, my take is `content:` is actually *specifically* for stuff that's not published yet; once it's on the website (and in that case it'll most likely be a full-fledged page coupled with .md/.pdf variants rather than *just* an md), it's a `feat:`

.github/workflows/deploy.yml
vzakharov, line 1: no, as discussed in the other thread (from previous code review)

docs/remove-before-merging/squash-message.md
```diff
@@ -29,20 +29,39 @@ attachment rather than a link. Its text is the operator's own, filled
...
+row and records that the list is local and extensible.
+
+**One call made inside that.** `content:` publishes: it is in the gate's regex
+alongside `feat:` and `fix:`. The prefix covers both content that reaches
+`public/` and drafts under `writing/` that are never built, and the subject line
+is all the gate reads, so one of the two has to be wrong. Publishing is the
+cheaper way to be wrong — a no-op republish nobody notices, against a live site
+silently stale behind `main`. The visible consequence is that this PR, which
+changes nothing built, spends a deploy on merge. Say so and the regex drops back
+to `feat|fix`, with `content:` under `public/` then deployed by hand.
```
vzakharov: regardless of the above comment, the squash message is not the right place for these implementation details

docs/remove-before-merging/squash-message.md
```diff
@@ -29,20 +29,39 @@ attachment rather than a link. Its text is the operator's own, filled
...
 `writing/` sits at the repo root deliberately: `public/` is served, and
 an unposted draft crawlable before it runs is the wrong order of
 operations, while `docs/` is where working artifacts go to be swept.
 
+`content:` is a new commit prefix, added because none of the existing
+ones describes a change to the site's own written material -- `docs:`
+is for documentation and `feat:` is for the code that serves it. It
+joins `feat:` and `fix:` in the deploy gate, since content that reaches
+`public/` changes what a visitor sees, and a missed deploy costs more
+than a redundant one.
```
vzakharov: no need to mention, it's a minor ride-along

writing/notes/the-five-percent.md
vzakharov, line 1: I like the idea of keeping that doc, to the point of making editing it mandatory after every code review session, so let's edit the place where we talk about them respectively. We must make it clear that it's not a permanent endeavor, rather than gathering material for a future article.

docs/remove-before-merging/squash-message.md
vzakharov: Taken, and it makes the prefix cleaner than the version I argued for: `content:` is written material the site does not yet serve, and a piece going live arrives as the page that serves it — `.md` and `.pdf` variants included — which is a `feat:`. So the trade I reasoned about does not exist, and the gate is back to `feat|fix` in 2c1108c. `CLAUDE.md`'s prefix row now says deploy-free by definition rather than leaving it to be inferred.

---

.github/workflows/deploy.yml
vzakharov: Reverted in 2c1108c — `.github/workflows/deploy.yml` is byte-identical to `main` again, gate regex and comments included.

---

docs/remove-before-merging/squash-message.md
vzakharov: Both notes gone in 2c1108c — the doc is the proposal and nothing else now. Recorded as a specimen too: the notes were true and traceable, and writing them was still the wrong call, because a squash body is read by someone scanning the log for what shipped.

---

docs/remove-before-merging/squash-message.md
vzakharov: Dropped in 2c1108c. The body now stops at where `writing/` sits and why; the prefix does not appear in it at all.

---

writing/notes/the-five-percent.md
vzakharov: Done in 2c1108c, in three places.

**Mandatory, and in the place review handling lives** — `CLAUDE.md` § "GitHub comments" now ends with it: a review session that changed something the agent had settled owes an entry, written before the session's last push rather than reconstructed later.

**Not permanent, stated where each half can be read on its own.** The rule paragraph and the file both name their end condition: the collection exists to make backlog row 18 draftable, and when that post runs the file and the rule retire together. `.claude/rules/writing.md` says the same for `notes/` generally — a notes file is scaffolding for one draft, not a document the repo keeps — and the row in `writing/linkedin/plan.md` says posting it is what retires them.

One thing I added beyond the ask, worth cutting if you disagree: the file also says that anything durable a review produces belongs in a rule or in `CLAUDE.md` **regardless** of whether it also lands here. Without that, a mandatory notes file is a plausible place for conventions to go and quietly stay.

This session's two specimens are in as the 7 September entry, one per existing family.

---

vzakharov: Restructured in 24f2d1e, on your follow-up: the top level is now the abstracted learnings, not the reviews. Each section carries `(×n)` — how many reviews bumped into it — and the file is sorted by that, so the post gets written from the top of the list rather than from whichever specimen is most vivid.

Six learnings over nine bumps. At `(×2)`: what it was handed treated as fixed; its reasoning written into the artifact; a published number read as verified. At `(×1)`: a form filled because it was a form; editing that removed the slack the voice is made of; hard-to-do mistaken for legible.

**Size, since you asked:** it was 7,933 bytes / 1,382 words / 142 lines; it is now 8,845 / 1,541 / 165 — the maintenance section is new and more than accounts for the growth, since the archaeology got shorter in the move. For scale, `.claude/rules/writing.md` is 5.5 KB and `plan.md` 18.7 KB.

The squeeze contract is written into the file: past ~400 lines, tighten first, then cut archaeology to the shortest account that still shows what the agent could not see, then drop learnings from the bottom — a learning cut for thinness comes back if something bumps into it again. ~400 lines is my guess at "too large", picked because `CLAUDE.md` already uses that number as its file rule of thumb; say a different one and it changes. Counts are by hand and stay that way, for the reason in the file: deciding that two reviews bumped into the same thing is the judgement the file is made of.

---

vzakharov:
### Attestation

Verified `bef2f0f` (`claude/linkedin-content-plan-5tzeoe`) against `main` @ `889e4fc` "fix: #28 re-window the commit chart's before/after rates (pr #29)".

- `./scripts/vet.sh` — **green**, twice: once after the base merge, and again on the final tree once the working-artifact sweeps had landed, so the SHA above is the one that was vetted rather than an ancestor of it.
- CI on this PR — **nothing runs on a PR in this repo** (`deploy.yml` triggers on push to `main` and `workflow_dispatch` only).
- CI-only buckets — **not dispatched**: this project has no `/test-on-gh` lane at all, not even a stub, and the diff is six markdown files that reach no credentialed service, no browser and nothing built.

**The base merge was not a no-op.** `main` brought in the #28 recomputation, which is the exact blocker backlog row 11 was parked in "Unsettled" for. Row 11 now carries the re-windowed figures (6.8 → 11.2 units of work a day, +65%, same median of about 380 changed lines either side), drops its ⚠, and its Unsettled entry goes with the objection it recorded — so Unsettled is three rows, not four. That is an editorial call made on the operator's behalf and is worth a look before merging; the dated specimen in `writing/notes/the-five-percent.md` was deliberately left as written, since it records what the review caught at the time.

Merging squashes under a `content:` subject, which is deploy-free by this branch's own definition — correct here, as nothing under `writing/` is built.

---

PR #29, opened 2026-09-06
fix: re-window the before/after rates on the commit-cumsum chart

### LINKED ISSUE #28: Re-window the before/after rates on the commit-cumsum chart
The commit-cumsum chart (`public/case-studies/assets/playgram-commit-cumsum.svg`, and the prose around it in `public/case-studies/playgram.md`) reports the switch into the cloud as **6.2 → 8.2 units of work per day**. Both windows are drawn wrong, so the ratio understates the effect.

Annotated chart: red arrow over the before-window, two green arrows bracketing the after-window with a days-off gap between them

## What is wrong

**The "after" window runs too long, and includes days that were not worked.**

- It should be measured between the two green lines in the annotation — the stretch that begins once work resumes after the break, not at the 25 April cloud switch itself.
- A considerable number of **days off** fall inside the current window. Idle days drag the daily rate down without saying anything about how fast the work went.
- It should **end at `4.1.0` (24 June)**. Before `4.1.0` the job was shipping the Bubble-parity app; after it, the job was bug fixes and later new features. That is a different kind of work, and averaging the two together measures the change of assignment rather than the change of tooling.

**The "before" window starts too early.** The red arrow currently begins at the very start of the project, which was a **docs-only phase**. Excluding documentation-only commits (as the chart already does) is not the same as excluding a period during which documentation was the work — those days land in the denominator as zeroes.

## What to do

1. Pick the four dates: start of the before-window (after the docs-only phase), its end; start of the after-window (after the break), and `4.1.0` as its end.
2. Recompute both rates over worked days, not calendar days.
3. Redraw the annotations on the chart to show the windows actually measured, so the picture and the number agree.
4. Update the prose in `playgram.md` and the chart's alt text, both of which currently quote 6.2 and 8.2.

## Why it matters beyond the chart

`writing/linkedin/plan.md` carries a post ("What parallelism actually measures as") whose entire content is this number. It is marked unsettled and blocked on this issue — it cannot be drafted until the windowing is fixed, because the post's whole claim is that the honest number is smaller than the marketing, and an honest number has to be honestly windowed.

---

vzakharov:
## Summary

- **Both published windows measured the wrong stretches.** The "before" window opened on 6 March, inside a phase whose commits were documentation rather than code — so it averaged the rate of writing code over days when no code was written. The "after" window ran 118 days to 21 August, four months past the point the job changed from building the app to maintaining it, and across a week of vacation. The resulting 6.2 → 8.2 understated the effect.
- **Two comparable windows replace them:** 11 March (where code commits overtake documentation in the Playgram repository) to 24 April, and 27 April (the first full day in the cloud) to 24 June (the `4.1.0` tag the chart already plots). That is 307 commits over 45 days and 607 over 54 — **6.8 and 11.2 units a day, a factor of 1.65** rather than the 1.34 the published pair implied.
- **Calendar days stay the basis**, which is the published unit and the slope of the line the reader is actually looking at. The single exclusion is the 6–10 May vacation, a known absence rather than a slow stretch; it is worth about 8% of the headline, so it is named in the prose and drawn as a labelled gap rather than left for the reader to infer.
- **The weekly panel agrees independently** — its median goes 44 → 74 over the same two windows, a factor of 1.68 against the upper panel's 1.65 — so its subtitle now reports the same two thirds instead of "about a third".
- **The annotation layer carries the claim:** a measurement bracket per window beneath the curve, the vacation as a gap in the second, and each rate label centred over its own bracket with the window dates in the sub-label, so the number can be checked against the segment it describes. `<desc>`, the alt text in all three cuts, and the mini/nano rate sentences are brought into line. The OG card and the three PDFs are re-rendered.
- **The full version gets the figure in prose, not only in alt text.** It reached that document's reader only through the chart's alt text and the labels inside the SVG, so it now lands in three places the reader passes through: a `Throughput` row in the summary table, the sentence in which the paragraph under the chart states the step-up, and the payoff of "Cloud-based agent VMs", the section whose 10/10 verdict this number is the measurement of. All three read it as **+65%** — the louder form of the same 1.65 — while the windows, the basis and the excluded week stay on the chart, in its labels and its `<desc>`, rather than being restated in prose. The mini and nano cuts already carried the pair in prose and are unchanged.

Rates were re-derived from the committed CSV — which is what the curve itself is drawn from — before any label was written.

## QA Checklist

- [ ] `brackets` — open `/case-studies/playgram` and look at the chart: a bracket under each window, the after-bracket in two segments with a "days off" gap between them, its right end directly under the `4.1.0` dot
- [ ] `labels` — each rate label sits centred over its own bracket and carries its window dates; nothing collides with the curve, the axis labels, or the weekly panel below
- [ ] `dark` — same chart in dark mode: brackets, gap and "days off" all legible against the dark plate
- [ ] `desc` — read the SVG's `<desc>` on its own and confirm it states the same two windows and two rates as the visible labels
- [ ] `alt` — the alt text is identical in all three cuts and matches what the chart now shows
- [ ] `prose` — the mini and nano rate sentences read naturally, quote 6.8 → 11.2, and each name the week off in May
- [ ] `full-prose` — in the full version, the `Throughput` table row, the paragraph under the chart and the paragraph before the cloud-VM verdict each carry the pair, agree with each other, and read in the author's voice
- [ ] `stale` — no `6.2`, `8.2`, `375` or `384` survives in a rate or unit-size context anywhere in the three cuts or the SVG
- [ ] `arith` — re-derive 6.8 and 11.2 from `playgram-commit-cumsum.csv` over the two windows and confirm they match every surface
- [ ] `og` — the social card at `playgram-commit-cumsum.og.png` shows the re-windowed chart
- [ ] `pdf` — the three PDFs carry both the new chart and the new prose

| `brackets` | manual-only | — | Bracket geometry against a curve is a visual judgment; the build renders the SVG but cannot see it |
| `labels` | manual-only | — | Label/curve collision is exactly what a hash check and a build both pass through silently |
| `dark` | manual-only | — | The dark palette is a `prefers-color-scheme` swap inside the SVG; needs a human eye per theme |
| `desc` | manual-only | — | Agreement between a description and a picture is a reading, not an assertion |
| `alt` | unit | ❌ | Three files' alt strings are byte-identical — assertable, but the repo has no content-lint suite to host it |
| `prose` | manual-only | — | Register and readability judgment |
| `full-prose` | manual-only | — | Same judgment, plus whether three spots in one document is emphasis or repetition |
| `stale` | unit | ❌ | A grep for retired figures in a rate context; same missing host as `alt` |
| `arith` | unit | ❌ | Pure function of the committed CSV — the one row here that genuinely belongs in `pnpm test`, but the literals live in a static SVG with no code to test |
| `og` | integration | ✅ | `pnpm content:og --check` hashes the source SVG against `og-renders.json`; wired into `vet.sh` |
| `pdf` | integration | ✅ | `pnpm content:pdf --check` hashes each document's whole source set against `pdf-renders.json`; wired into `vet.sh` |

Fixes #28

https://claude.ai/code/session_01PiHtCmbkQKj5hTCHciEg5M


vzakharov:
Proposed squash title/body:

```
fix: #28 re-window the commit chart's before/after rates (pr #29)
```

```
The chart reported the switch into the cloud as 6.2 to 8.2 units of
work per day, and both windows behind those figures measured the
wrong stretch. The "before" window opened on the first commit, inside
a phase whose commits were documentation rather than code, so it
averaged the rate of writing code over days when no code was written.
The "after" window ran 118 days to the handover -- four months past
the point the job changed from building the app to maintaining it,
and across a week of vacation.

Measure two comparable windows instead: 11 March, where code commits
overtake documentation in the Playgram repository, to 24 April; and
27 April, the first full day in the cloud, to the 4.1.0 tag on 24
June. That is 307 commits over 45 days against 607 over 54 -- 6.8 and
11.2 a day, a factor of 1.65 rather than the 1.34 the published pair
implied. Calendar days stay the basis, being the published unit and
the slope of the line the reader is actually looking at, and the one
exclusion is the 6-10 May vacation: a known absence rather than a
slow stretch, and worth about 8% of the headline. The weekly panel
agrees independently, its median going 44 to 74 over the same two
windows, so its subtitle reports that same two thirds. The median
unit size becomes "about 380 changed lines" either side, the
published 375/384 pair having been measured across the switch and
inviting a comparison of two figures whose point is that they match.

The annotation layer now carries the claim rather than floating free:
a measurement bracket per window beneath the curve, the vacation as a
labelled gap in the second, and each rate label centred over its own
bracket with the window dates in its sub-label, so a reader can check
the number against the segment it describes. The full version stops
leaving the headline to the chart and its alt text, too -- as +65%,
the louder form of the same 1.65, the pair reaches its summary table,
the paragraph that walks the chart, and the payoff of the cloud-VM
section whose verdict it measures. Nothing here shares a runtime with
anything else: the SVG's labels and <desc>, the alt text triplicated
across the three markdown cuts the content pipeline serves raw at
their own URLs, and the rate prose in each of them all change
together by hand. Editing the SVG stales the Open Graph card and all
three PDFs, editing a document stales its own, and vet.sh gates both
on hashes only.

Fixes #28

```

---

vzakharov:
### Attestation

Verified `18d87c2` (`claude/28-rewindow-commit-rates-89dv6q`) against `main` @ `ee0a55e` "feat: serve .md and .pdf at each document's own URL (pr #25)".

- `./scripts/vet.sh` — **green** (including `content:og --check` and `content:pdf --check`, which hash the re-rendered card and the three PDFs against their sources)
- CI on this PR — **nothing runs on a PR in this repo** (`deploy.yml` triggers on push to `main` and `workflow_dispatch` only)
- CI-only buckets — **not dispatched: `/test-on-gh` is not present in this project**, and the diff reaches no surface needing credentials, real services or a browser — it is the committed SVG, three markdown cuts and their re-rendered derivatives

`origin/main` was still at `ee0a55e` at the merge check, and the branch already contained it, so no base merge was needed.

The vet run does not judge the change: it hashes the derivatives and builds every page, but it cannot see whether the brackets, the labelled vacation gap and the rate labels read correctly against the curve in either theme. The `## QA Checklist` in the PR body is where those rows live, and the nine `manual-only` ones are still open.

---

PR #31, opened 2026-09-08
feat: reposition the site and CV around hands-on CTO for AI-native delivery

vzakharov:
## Summary

- **The CV comes in two framings.** `/{locale}/cv` renders a *fractional hands-on CTO for AI-native delivery* variant in place — no redirect, since a static export has no server-side one — and `/{locale}/cv/cto` is the self-describing address for the same content, declaring `/cv` canonical so the pair does not split link equity. `/{locale}/cv/dev` keeps the developer framing, unchanged in substance, at its own URL, and a visible CTO/Dev switch sits beside the locale picker.
- **Only differing prose exists twice.** `cv.variants.cto` carries `metadata`, `header.tagline` and `profile`; `cvMessages` merges it over the base catalogue a section at a time, so a key the variant does not restate cannot drift. Experience, tech stack, education and contact are one set of facts stated once — and `whatIOffer` needed no override at all: its subsections became keyed blocks with the per-variant order in code, the pattern `experience` and `techStack` already use.
- **The home page is dev-first.** Writing and music each move to a page of their own, reached from a *See also* footer line; the dev intro states what is on offer instead of that its author is looking for work; the case-studies block folds into a four-card featured grid, with the earlier three projects named on one line; Professional Work becomes Recent Work with Playgram at the top and Voicemod dropped.
- **Nothing retitles history.** Playgram is "Developer" in both variants — the developer identity is the CTO claim's evidence, which is why both CVs exist rather than one replacing the other. Drive-by fixes: `ru.json` said "с 2019 года" where `en.json` says 2020, and the CV called `almostmagic` a Python package.
- Plan: `docs/plans/cto-positioning.completed.md`.

## QA Checklist

- [ ] `default-framing` — open `/en/cv`: tagline reads *Fractional hands-on CTO for AI-native delivery*, the profile opens on the numbers, and What I Offer lists Engagements / The Engineering System / AI Expertise / Working Style.
- [ ] `dev-framing` — click `Dev`: the URL becomes `/en/cv/dev` and the page returns to today's developer tagline, profile and three capability bullets.
- [ ] `switch-composes` — from `/ru/cv/dev`, flip the locale then the variant, and the reverse; each control preserves the other's choice.
- [ ] `canonical` — view source on `/en/cv/cto`: `<link rel="canonical">` points at `/en/cv`. `/en/cv` and `/en/cv/dev` carry none.
[...]

PR #34, opened 2026-09-09
chore: one job per loop skill, and a size cap on squash bodies

vzakharov:
## Summary

- **Two source commits triaged, ported as five rationales.** `scripts/check-squash-message.sh` measures the squash proposal against the caps `/squash-message` states — an 80-char one-line title, a body of at most 50 lines at 72 wide — and joins the vet fan-out as its tenth check, reading only the proposal so it overlaps the other nine safely.
- **`/plan` publishes its own plan as a draft PR**, so a plan is reviewed as a diff with inline comments and threads rather than as chat prose. The trigger sits in `/plan` because that is where the plan becomes pushed, so every entry into planning gets a PR.
- **`/pr` loses its task argument and gains three modes over the PR object** — plan-open, create, and refresh. The old duplicate-PR guard inverts: an existing PR is the expected state at the end of implementation, so it gets its body and QA checklist re-derived against the real diff the plan could only forecast.
- **`/implement` becomes `/go`**, converging the command with the vocabulary `/plan`'s approval gate already listens for. The retired name is declined rather than kept resolvable by a redirect stub. `/handle` reads the plan and review lanes against one predicate — implementation has begun iff `docs/plans/` holds a file that is not a draft — so a draft plan plus unanswered feedback reads as review *of the plan*, and no review comment ever unlocks source edits.

### Triage

|---|---|---|
| `367f0bb` squash-message size caps | **take** + **translate** | Script and skill edits apply verbatim — this repo carries the same proposal paths. The `vet.sh` wiring is translated: local `vet.sh` is a `rewrite`, so the check joins the existing `run-parallel.sh` fan-out rather than a bare serial call. Its CLAUDE.md hunks are **skip (stack-bound)** — they concern the boilerplate's stub `vet.sh` exit code and stub-skill roster, neither of which exists here. |
| `4998149` give each loop skill one thing to own | **take** | All nine touched skills are adopted here, so the restructuring applies as-is. `.claude/skills/audit-github-backlog/` hunks are **skip (not adopted)** — declined here, reason still holds. `README.md` and `docs/catalog.md` hunks are **skip (not adopted)** — this repo's README is its own, and the catalog is declined `never`. |

Three Step-4a decisions, all recorded in `source.json` so none comes back: `.claude/skills/go/` and `scripts/check-squash-message.sh` adopted; `.claude/skills/implement/` **declined** on the operator's call — a permanent extra row is not worth keeping a retired name resolvable, and an older handoff block saying `/implement` is a word an agent reads rather than a skill it must load. Recorded as `never`, since a redirect stub only ever earns its place in the window right after a rename.

Re-reading the declined reasons found one that had stopped holding: `/test-on-gh` was declined for having no test suite, and one now exists. The skill still does not apply, but for the reason that actually gates it — nothing runs on pull requests here, and the suite is fast enough to sit in the vet fan-out — so the reason is restated in those terms with the condition that would reopen it named.

## QA Checklist

- [ ] `caps-pass` — `scripts/check-squash-message.sh` on a branch with no proposal reports "nothing to measure" and exits 0.
- [ ] `caps-fail` — point it at a file whose title block is >80 chars or whose body exceeds 50 lines / 72 cols; confirm it names each value against its cap and lists over-width lines by number.
- [ ] `caps-exempt` — a body line that is a single long token (a URL) does not trip the width check.
- [ ] `vet-wired` — `./scripts/vet.sh` runs green with `squash` in the fan-out, and `run-parallel.sh` reports it among the checks.
- [ ] `go-resolves` — `/go` loads and carries the whole procedure; no skill directory named `implement` remains.
- [ ] `closure` — `bash scripts/check-skill-catalog.sh` passes: every `@.claude/skills/…` pointer resolves and no stub is unhydrated.
- [ ] `no-stale-refs` — nothing in `.claude/` or `CLAUDE.md` still cites `/pr` Step 1a/1b, the deleted plan gate, or `@.claude/skills/implement/SKILL.md` as the place a procedure lives.
- [ ] `watermark` — `source.json` parses, `lastSyncedSha` is `4998149…`, and the two new adopted paths are present.

| `caps-pass` | yes | no | Exercised incidentally by the vet run; no unit test. |
| `caps-fail` | yes | no | The natural first candidate for `scripts/*.test.ts`, matching `type-overlap-check.test.ts`'s pattern. |
| `caps-exempt` | yes | no | Same. |
| `vet-wired` | yes | yes | `./scripts/vet.sh` — run green on this branch. |
| `go-resolves` | partly | yes | The directory's absence and pointer closure are checked by `scripts/check-skill-catalog.sh`; that `/go` reads correctly on invocation is not. |
| `closure` | yes | yes | `scripts/check-skill-catalog.sh` — run green on this branch. |
| `no-stale-refs` | partly | no | Grep-checked by hand this session; no checker verifies that a citation still *means* what it meant. |
| `watermark` | yes | no | JSON validated by hand; nothing asserts the SHA advanced. |

Prose changes only — no page, route, component or style is touched, so the built site is unchanged and this does not need a deploy. `chore:` as the squash subject keeps the deploy gate closed deliberately.

https://claude.ai/code/session_013AapExDuqdBf6ENd1fUH2z


vzakharov:
Proposed squash title/body:

```
chore: one job per loop skill, and a size cap on squash bodies (pr #34)
```

```
The boilerplate this repo vendors its agent infrastructure from kept
editing it, and two commits had landed there since the last sync. Both
apply here: every skill they touch is adopted, and neither is bound to
the source's stack.

`scripts/check-squash-message.sh` settles the part of the
squash-message target a machine can. `/squash-message` states it as
prose — a one-line title, a body hard-wrapped at ~72 — and its
tighten pass is an agent reading its own output; the script measures
the proposal's two fenced blocks against 80 chars of title and 50
lines of body at 72 wide. It joins the vet fan-out as its tenth check,
reading only the proposal, so it catches what authorship cannot: a
proposal edited by hand, or outgrown by a later base merge.

The rest gives each skill in the loop one thing to own. `/plan` now
publishes its own plan as a draft PR, so a plan is reviewed as a diff,
with inline comments and threads, instead of chat prose; the trigger
sits in `/plan` because that is where a plan becomes pushed. That leaves
`/pr` with no reason to take a task argument: it owns the PR object in
three modes, the old duplicate-PR guard inverted into a refresh that
re-derives body and QA checklist against the diff the plan could only
forecast. `/implement` becomes `/go`, the retired name declined rather
than kept resolvable by a stub. One predicate carries the review loop
— implementation has begun iff `docs/plans/` holds a file that is not
a draft — so a draft plan plus unanswered feedback reads as review of
the plan, and no review comment unlocks source edits; only an invocation
does.

```

---

vzakharov:
### Attestation

Verified `3da0662` (`claude/sync-boilerplate-njt3gz`) against `main` @ `3cc95b1` "feat: reposition the site and CV around hands-on-CTO delivery (pr #31)".

- `./scripts/vet.sh` — **green**, re-run at this head after the base merge and the squash retitle below. Post-sweep, this PR's own new check resolves through its ladder to the deleting commit: `ok — docs/remove-before-merging/squash-message.md (swept in 3da0662, read from history) (title 71/80, body 29/50 lines, widest 72/72)`.
- CI on this PR — **nothing runs on a PR in this repo**. `deploy.yml` triggers on push to `main` and `workflow_dispatch` only.
- CI-only buckets — **not dispatched: `/test-on-gh` is not adopted here**, and the diff reaches no surface needing credentials, services or a browser.

**Re-verified after `main` advanced** `4068aee` → `3cc95b1`. That advance is not independent of this branch: it edits `.claude/skills/squash-message/SKILL.md`, which this branch rewrites. Merged as `efa0df9`, resolving one real conflict in Step 3 — the base's "a re-run rewrites; it never accretes" paragraph and this branch's `check-squash-message.sh` measure step are complementary, so both are kept, and the paragraph the two sides re-wrapped from opposite ends was re-flowed.

The same merge made this PR's **own** squash proposal non-conforming: the base added a rule that a trap for whoever edits the area next is not part of the permanent record, and the body's account of the script's proposal-finding ladder was exactly that. Cut in `3ad117d` — the script's header already carries it — taking the body from 34 lines to 29.

**Squash title retitled** in `4e3b832`, on the operator's call: the old subject named the sync's provenance (`sync the agent boilerplate forward to 4998149`) rather than what changed. It now names both strands the body carries. This PR's title is set to match, since GitHub prefills the squash subject from it.

`main` @ `3cc95b1` as of this check.

---

PR #36, opened 2026-09-09
feat(cv): refresh the CV and rebuild the theme picker as two-state

vzakharov:
## Summary

- **Retitles the CV's experience entries up to `randddb`** — Playgram → **Fractional CTO**, `englishForKids`/`orcool` → **Senior Fullstack Developer**, `randddb` → **Fullstack Developer** — and drops the `– Project Work` qualifiers, since most of the history the CV now shows _is_ project work. The Russian counterparts leave the seniority and stack terms untranslated.
- **Reworks the CV's header**: the theme toggle moves into its top-right corner, and an `[EN] [RU]` chip pair with a `.pdf` link sits _below_ the hero — the shape the article header already uses for a document's cuts and files, since that row acts on the document rather than introducing it. The flag picker and the `globalThis.print()` button both go, and the name heading links home.
- **Rebuilds the theme picker as a two-state control.** Light and dark are the only states a reader sees, over one rule: **a pick that agrees with the reader's system scheme is stored as no preference at all.** So the site follows the system until the reader picks the other scheme; that pick then stands — through the system's own changes included — until they pick their system's scheme back, which hands control to the system again. `preferredColorScheme` is the whole rule and has its own tests; the click reads the system scheme from the media query rather than `useComputedColorScheme`, whose first render reports `light` whatever the OS says. Both icons render and CSS picks one off `data-mantine-color-scheme`, so the right icon is painted on the first frame with no placeholder — and the toggle carries `print-hidden` itself, which four of `main`'s six call sites did not.
- **Parks the toggle in the header's corner, and takes its chrome off.** On home and the CV it cost a row of its own — a 38px control plus the page stack's gap — for something that needs no band; it now sits in the top-right of the header itself, out of the flow, through a `CornerHeader` in `shared/ui` that owns both halves of the positioning so a caller cannot pair an absolute corner with a header that forgot to be its context. Everywhere it renders it also drops `variant="default"`, the hairline box the site's real controls wear, for a bare grey glyph that comes up to full on hover.
- **Brings the CV into the committed-PDF pipeline** at `public/cv/<variant>/<locale>.pdf`, keyed off the canonical address so the short rungs offer the same file rather than duplicating it. `content:pdf` runs under `tsx` now and its source list is factored into `PRINT_SOURCES` / `DOCUMENT_SOURCES` / `CV_SOURCES`; a CV printable hashes only its own locale's catalogue. All seven PDFs are re-rendered and committed.
- **Four extractions.** `ChipNav` and `FileLink` move down to `shared/ui` (`pages/cv` may not reach sideways into `pages/case-studies`) and `DocumentFile` to `shared/typings`, as the plan called for; beyond it, `pageFile` factors the download-name shaping into `shared/config`, where a client component can also reach it, and the hover dim becomes `hoverDim` — a `shared/ui` module class handed out through the public API, since the article's back-link claims it as well as `FileLink`.
- **Holds `writing/notes/` to its own line ceiling.** The rule — "past 400 lines, squeeze instead of growing" — was a bullet inside the file it governs, so nothing read it on the way to appending a section and the file reached 447 lines. `scripts/check-notes-length.sh` joins vet's fan-out and fails on it; the notes themselves are squeezed back to 397 in the order the file prescribes, one thin learning dropped off the bottom.

## QA Checklist

- [ ] `titles` — open `/cv/cto/en` and `/cv/cto/ru`; each experience entry shows its new role title, and the company and dates still read off the line below it
- [ ] `lang-chips` — on the CV, `EN` and `RU` render as chips below the hero, the current one inverted and inert; clicking the other lands on the same framing in the other language
- [ ] `cut-chips` — open `/case-studies/playgram`; the Full / Mini / Nano switcher looks and behaves exactly as it did before the chip control moved to `shared/ui`
- [ ] `pdf-link` — the CV's `.pdf` link downloads (rather than opens) `vova.cv.cto.en.pdf`, and the file is the CV as the print stylesheet renders it, with no header controls on it
- [ ] `pdf-rungs` — `/cv` and `/cv/cto` offer the same PDF their canonical address does, not a 404
- [ ] `home-link` — clicking the "Vova Zakharov" heading on the CV goes to `/`, and the heading still reads as a heading (no underline, inherited colour)
- [ ] `hover-dim` — the article's `← Case studies` back-link and the CV's `.pdf` link both still dim on hover, and `globals.scss` carries no `.hover-dim` for them to have fallen back to
- [ ] `two-state` — on a dark system with nothing stored: one click gives light and `mantine-color-scheme-value` holds `light`; a second click gives dark and **the key is gone**, not set to `dark`. Mirrored on a light system
- [ ] `override-sticks` — with `light` stored, flip the OS to light and back to dark: the page stays light throughout and the key is untouched
- [ ] `system-theme` — with nothing stored, flip the OS between light and dark with the site open: every page follows, live
- [ ] `icon-frame` — hard-load on a dark system: the sun is correct on the first frame, with no empty box and no moon-to-sun pop before it
- [ ] `toggle-everywhere` — home, writing, music, the case-studies index, an article and the CV each render the toggle, and none of them prints it
- [ ] `toggle-corner` — on `/` and `/cv/cto/en` the toggle sits in the top-right of the header — level with the avatar, level with the name — and no empty band is left above either hero
- [ ] `toggle-quiet` — in both schemes the toggle is a borderless grey glyph, and hovering it brings it to full strength rather than dimming it
- [ ] `print` — Ctrl+P on the CV still produces the same sheet the committed PDF holds
- [ ] `notes-ceiling` — `scripts/check-notes-length.sh` passes on the branch, and fails naming the file and the overage when a line is appended to `writing/notes/the-five-percent.md`

| Item                | Automatable | Covered? | Notes                                                                          |
| ------------------- | ----------- | -------- | ------------------------------------------------------------------------------ |
| `titles`            | unit        | ❌       | Catalogue assertion over `cv.experience.*.title` for both locales              |
| `lang-chips`        | e2e         | ❌       | Drive `/cv/cto/en`, click `RU`, assert the URL and the inverted chip           |
| `cut-chips`         | unit        | ❌       | Component test over `ChipNav`: current renders inert, others as links          |
| `pdf-link`          | integration | ❌       | Assert the anchor's `href`/`download` and that the file exists under `public/` |
| `pdf-rungs`         | integration | ❌       | Same assertion over the two short rungs, against the canonical path            |
| `home-link`         | e2e         | ❌       | Click the `h1` on `/cv`, assert navigation to `/`                              |
| `hover-dim`         | integration | ❌       | Assert both anchors carry the module's hashed class in the prerendered HTML     |
| `two-state`         | unit        | ✅       | `preferredColorScheme` — the rule as a table, both systems both directions      |
| `override-sticks`   | e2e         | ❌       | Emulate `prefers-color-scheme`, click, flip the emulation, assert storage       |
| `system-theme`      | e2e         | ❌       | Emulate `prefers-color-scheme` and assert `data-mantine-color-scheme`          |
| `icon-frame`        | integration | ❌       | Assert both icons and their scheme classes are in the prerendered HTML          |
| `toggle-everywhere` | e2e         | ❌       | Assert the control renders on each of the six routes and is hidden in print     |
| `toggle-corner`     | e2e         | ❌       | Assert the control's box sits inside the header's and no row precedes it       |
| `toggle-quiet`      | manual-only | —        | Resting opacity and the hover direction are a visual judgment                  |
| `print`             | manual-only | —        | Print fidelity is a visual judgment against the committed PDF                  |
| `notes-ceiling`     | —           | ✅       | The row _is_ a vet check; `vet.sh` runs it on every branch from here            |

`two-state` is the one behavioural row the suite covers: `preferredColorScheme` is a pure function of two arguments, which is what CLAUDE.md § "Testing" says belongs in the runner rather than a checklist row. `notes-ceiling` is covered differently — it is itself the check, so vet asserts it rather than a test. The rest of the toggle's behaviour was driven for real in an earlier session — clicks and `prefers-color-scheme` emulation over CDP, confirming every storage transition above — but that harness was scratch, not a committed test. For everything else `pnpm build` plus the two render `--check`s are what stand behind this branch.

**Review round (2026-09-09).** Four threads, all replied to and left open. Two produced code: the hover dim moved off `globals.scss` into a `shared/ui` module class, and the `writing/notes/` ceiling gained the vet check that the file's own prose could not enforce. Two were deferred by the operator and filed instead — #39 (re-API `cx` to take the module first, then sweep) and #40 (the theme toggle should come from a layout rather than from every page that wants one). `CornerHeader` stays as-is here, per "but not here".

https://claude.ai/code/session_01HGV2q91etXHLc8qK2G6bba


vzakharov:
Proposed squash title/body:

```
feat: refresh the CV's roles and controls, rework the theme picker (pr #36)
```

```
The CV filed five years of work under one word, "Developer", including
the engagement it now leads with. Its header carried three controls,
each behaving unlike anything else on the site: a flag showing the
language you were already reading, a button that opened the browser's
print dialog, and a theme toggle cycling three states.

The titles now say what the work was — Fractional CTO at Playgram,
Senior Fullstack Developer for the two project engagements above it,
Fullstack Developer for randddb — in both catalogues, the seniority
and stack terms left untranslated in the Russian one as they are in
the market it reads to. The language switch becomes the chip pair the
article header already offers for a document's cuts, both languages
shown and the current one inert; the print button becomes the `.pdf`
link every other document offers. Both move down to `shared/ui`, the CV
having no sideways reach into `pages/case-studies`, and sit below the
hero as the article's row does, since they act on the document rather
than introduce it. The CV joins the committed-render pipeline — one
PDF per canonical `/cv/<variant>/<locale>`, the short rungs linking to
theirs — and `content:pdf` moves to `tsx` to reach those addresses,
its source list split so an English reword leaves the Russian sheet
alone. The name heading links home.

The theme toggle keeps two visible states over one rule: a pick that
agrees with the reader's system scheme is stored as no preference at
all. So the system leads until the reader picks the other scheme; that
pick then stands, through the system's own changes included, until they
pick the system's scheme back, which hands control to it again. The
click reads the media query rather than a hook whose first render says
`light` whatever the OS says, and both icons render for CSS to choose
between, so the right one is painted on the first frame rather than
after hydration.

It also stops presenting as a control. Where the site's buttons and
chips wear a hairline box, the toggle is a bare grey glyph that comes up
to full on hover, and on home and the CV it sits in the header's
top-right corner rather than on a row of its own, so a page that spent
a band on it now starts at its heading. `CornerHeader` owns both halves
of that positioning, so no caller can pair an absolute corner with a
header that forgot to be its context. Riding along: `writing/notes/`
gets the line ceiling it previously only stated in prose, as a vet
check.

Claude-Session: https://claude.ai/code/session_01HGV2q91etXHLc8qK2G6bba
```

---

src/pages/case-studies/ui/article-page.tsx
```diff
@@ -93,7 +93,7 @@ export async function ArticlePage({ params }: Props) {
             <InternalLink
               href={collectionRoute(COLLECTION)}
               size="sm"
-              className={classes['hoverDim']}
+              className="hover-dim"
```
vzakharov: hmmm, why are we bypassing modules here? not a fan (in other such places too)

src/shared/ui/chip-nav.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function ChipNav({ chips }: ChipNavProps) {
+  return (
+    <Group component="nav" gap={8} wrap="wrap" fz="sm" className="print-hidden">
+      {chips.map(({ label, href, current, hrefLang }) =>
+        current ? (
+          <Box
+            key={label}
+            component="span"
+            aria-current="page"
+            className={cx(classes['chip'], classes['chipCurrent'])}
```
vzakharov, line 34: let's file an issue to re-api cx to a la `cx(classes, 'chip', 'chipCurrent')` & sweep

writing/notes/the-five-percent.md
vzakharov, line 1: isn't 400 lines the agreed ceiling? let's put a vet.sh-controlled check

src/shared/ui/corner-header.tsx
vzakharov, line 1: frankly, not a fan of idea of putting this in every consumer that needs a theme switcher -- because all of them do. I'd rather prefer a layout (if that's the right term)-based approach where the theme always comes. but not here, so let's file an issue

src/pages/case-studies/ui/article-page.tsx
vzakharov: Fair — reached for a global because `print-hidden` was sitting next to it, without asking what made *that* one global. It has a reason that doesn't transfer: the content pipeline emits `print-hidden` into HTML strings (`rehype-media-embeds.ts`), where a hashed module class is unreachable. Nothing about the hover dim needs that.

So it's back in a module and crosses the slice boundary the way every other export does — `hoverDim` from `shared/ui`'s public API, same shape `theme.module.scss` already uses to hand Mantine its class names. `globals.scss` is down to the colour tokens. 0db9273.

The other such places are all `print-hidden`, which I've left global for the reason above — a module class there would have to be a literal name anyway, so the hash is what makes it impossible rather than the style. Happy to sweep the TSX call sites onto an exported constant if you want the typo protection, but that wouldn't restore modules.

---

src/shared/ui/chip-nav.tsx
vzakharov: Filed: https://github.com/vzakharov/vovazakharov.com/issues/39.

Wrote up the one thing that needs deciding before the sweep, since every current call site mixes a module key with a global utility or a passed-through `className` (`cx('print-hidden', classes['pageFooter'])`, `cx(classes['header'], className)`), so the new signature has to say what happens to those. Three options in the issue, no recommendation — the call sites should decide it. Also noted the part that makes it worth doing beyond the repetition: the keys can be typed against the module's own keys, so a renamed class fails the type check instead of rendering `undefined`.

---

writing/notes/the-five-percent.md
vzakharov: Right, and it's this file's own rule rather than the repo's — `CLAUDE.md` says ~450 for *production* files, while "How this file is kept" says "past ~400 lines, squeeze instead of growing". The file was at 447, and the append that broke the rule was reading it on the way in.

Both halves done:

- **The check** — `scripts/check-notes-length.sh`, in vet's fan-out (9bd14f9). POSIX `sh` and `wc` over `writing/notes/*.md`, which nothing else there touches. It reports and stops rather than trimming, since which of the three squeezes applies is a judgement.
- **The squeeze** — 447 → 397 (6e62e4a), in the order the file prescribes. Prose, then archaeology (most of it — every bump is now the shortest account that still shows what the agent couldn't see), then one learning off the bottom: "Editing removes slack, and the slack was the voice" was the thinnest section at one bump and three lines of evidence, and it comes back if something bumps into it again.

That's with this review's two bumps added on the way — the global class read off the globals beside it, and this one: a constraint the agent stated in prose where a mechanism was available, which is the section it landed in.

---

src/shared/ui/corner-header.tsx
vzakharov: Filed: https://github.com/vzakharov/vovazakharov.com/issues/40. Left `CornerHeader` as-is on this branch, per "not here".

Agreed on the objection — all six pages render the toggle themselves, so it isn't an opt-in in any useful sense; a page that forgets it ships without one and nothing says so. (That already happened in the smaller form: four of the six `<Group>` sites were missing `print-hidden` before this branch put it on the control.)

Two things the issue flags for whoever picks it up. The corner is currently positioned against the *header's* box, which is why `CornerHeader` owns both halves of it — from a layout the anchor becomes the viewport or the page shell, so on home and the CV the control stops being level with the avatar and the name, which wants `/preview` rather than a decision from source. And the article's back-link row is a `justify="space-between"` nav whose right-hand child is the toggle, so hoisting it leaves a `space-between` with one child.

---

vzakharov:
### Attestation

Verified `ba80f81` (`claude/cv-refresh-rbfhk2`) against `main` @ `754b2af` "feat: carry the CTO positioning to the landing page and card (pr #35)".

- `./scripts/vet.sh` — **green** (re-run at `ba80f81` itself, after the working-artifact sweeps, so this names the head rather than an earlier commit)
- CI-only buckets — **`/test-on-gh` is not present in this project**, so none was dispatched

`main` was already contained in the branch at `754b2af`, so step 2's merge was a no-op and the vet run covers the merged state directly. `/check-merge` reported `contained` against that same tip. `docs/pr/36/`, both `docs/plans/*.completed.md` and `docs/remove-before-merging/` are swept — `docs/` no longer exists at this head, and `main` never carried anything under it.

---

PR #41, opened 2026-09-10
feat: the theme toggle comes from the layout

### LINKED ISSUE #40: The theme toggle should come from a layout, not from every page that wants one
Raised in review on [#36](https://github.com/vzakharov/vovazakharov.com/pull/36):

> frankly, not a fan of idea of putting this in every consumer that needs a theme switcher -- because all of them do. I'd rather prefer a layout (if that's the right term)-based approach where the theme always comes. but not here, so let's file an issue

## What

Six pages render `<ThemeToggle />` themselves, two of them through `CornerHeader`:

| Page | How it renders the toggle |
| --- | --- |
| `pages/home` | `CornerHeader corner={<ThemeToggle />}` |
| `pages/cv` | `CornerHeader corner={<ThemeToggle />}` |
| `pages/writing` | a `<Group>` of its own |
| `pages/music` | a `<Group>` of its own |
| `pages/case-studies` (index) | a `<Group>` of its own |
| `pages/case-studies` (article) | a `<Group>` beside the back-link |

That is every page on the site, which is the objection: an opt-in every consumer takes is not an opt-in. A page that forgets it ships without a toggle and nothing says so, and the four `<Group>` sites are also where `print-hidden` went missing before #36 put it on the control itself.

## The shape to reach for

The toggle moves into the app layer — `src/app/ui/root-layout.tsx`, or a wrapper it renders — so it lands once, out of the page's flow, and a page renders nothing for it. `CornerHeader` and its two call sites go away with it; the four `<Group>`s lose a child.

Two things to work out on the way:

- **Where it sits per page.** The current corner is positioned against the *header's* box, which is why `CornerHeader` owns both halves of it. From a layout the anchor is the viewport or the page shell instead, so the control's position stops tracking the header — fine on the pages that put it in a row today, a visible change on home and the CV, where it is currently level with the avatar and the name. Worth looking at with `/preview` rather than deciding from source.
- **The article's back-link row.** On `/case-studies/<slug>` the toggle is the right-hand end of a `justify="space-between"` nav whose left is the back-link. Hoisting the toggle leaves that row with one child and a `space-between` that no longer means anything.

Neither is load-bearing on #36, which is why this is filed rather than done there.

vzakharov:
## Summary

- **The theme toggle comes from the layout.** Six pages rendered `<ThemeToggle />` themselves — an opt-in every consumer takes, which is the objection in #40. A new `src/app/ui/theme-corner.tsx` gives it one home: `RootLayout` renders it once inside `<ThemeProvider>`, absolutely positioned against the initial containing block, so it lands at the page's corner regardless of a sheet's own max width and scrolls away with the top of the document. Every page now renders nothing for it.
- **`CornerHeader` retires** rather than losing its `corner` prop: with the corner hoisted it wraps a `<header>` and nothing else. Home and the CV spell that element themselves; `/writing`, `/music` and `/case-studies` lose a whole `<Group>` row, and the article's nav keeps its back-link without `justify="space-between"`.
- **The CV's locale / `.pdf` row becomes a tinted strip**, replacing the header's hairline on screen — a rule immediately above a filled band separates twice. `.header` keeps its rule inside `@media print`, where the strip is gone. Its inline padding is the Card's 24px, so the chips line up with the copy in the sheets below.
- **A chip's hover tints instead of filling.** `.chipLink`'s hover copied `.chipCurrent` exactly, so hovering `RU` painted it the same solid black as the selected `EN`. Hover takes `surface-strong` and keeps the chip's text and border, leaving solid foreground to mean only *current*. `ChipNav` is shared, so the case-study cut chips get it too.
- **Both tints needed the site's first filled surface**, so `CSS_COLORS` gains a `surface` / `surface-strong` pair on the model of the `border-hairline` family — the strip takes the first, and the hover needs a tint legible *on* it, which one token cannot be. `pnpm styles:codegen` regenerated `styles/_tokens.scss`, and all three palettes pass the new values.

**Two things `/preview` changed from the plan.** The strip has no `border-radius`: `defaultRadius` is `0` and every Card is square, so 8px made it the one rounded thing on the page. And at a **true** 360px viewport the CV's centred title ran 22px under the corner toggle — in every variant and both locales — so the CV's header takes a `padding-top` below `sm`. The clearance is the header's because the corner is every page's and that collision is not.

All seven committed PDFs re-render (`PRINT_SOURCES` hashes `src/shared/ui` wholesale). **Their visible ink is unchanged** — verified by comparing the glyph-show and rectangle operators of each against `main`, normalised of font-subset numbering. What moves in the CV's stream is paint order: its header was `position: relative` only to anchor the corner.

`docs/issue/40/issue.md` is a `prettier --write` only — it landed unformatted with the export in b8e0c7d and was failing `format:check` before this branch's work.

### Also: three corrections to `/preview`

Running the skill for this change disproved three of its own claims, so they change with it: its documented 500px capture floor is escapable via CDP `Emulation.setDeviceMetricsOverride` (which is what caught the CV collision); **a `:hover` rule is invisible in a capture and silently so**, because headless Chromium reports `(hover: none)` and the `hover` mixin takes its `:active` branch; and the theme is `data-mantine-color-scheme`, not next-themes' `class="dark"`, so its verification step could never confirm what it checked.

## QA Checklist

- [ ] `once` — every page (`/`, `/cv`, `/writing`, `/music`, `/case-studies`, `/case-studies/<slug>`) shows exactly one theme toggle, and no page component imports `ThemeToggle`
- [ ] `corner` — the toggle sits in the page's top-right corner at the same inset on all six pages (32px, 48px above `sm`), clear of the first line of content and reading as a corner rather than a stray icon on a wide desktop
- [ ] `themes` — clicking it still flips light ↔ dark on every page, and the icon shows the scheme a click gets you
- [ ] `cv-phone` — on a **real** phone (or true viewport emulation, not `--window-size`), `/cv` in both variants and both locales shows the centred title clear of the toggle; the plain 500px-floor capture cannot show this
- [ ] `lift` — `/writing`, `/music` and `/case-studies` lost a row: their first content sits ~86px higher and nothing collides
- [ ] `article-nav` — on `/case-studies/<slug>` the back-link stays at the left of its nav row, with no orphaned `space-between` gap
- [ ] `toolbar` — on `/cv` the EN/RU chips and the `.pdf` link sit on a square tinted strip where the hairline used to be, its left edge lining up with the card copy below; the tint reads as a surface in light **and** dark
- [ ] `chip-hover` — hovering the unselected locale on `/cv` and a cut chip on `/case-studies/<slug>` tints it without filling it, keeping its border, in both schemes and on both the strip's ground and the page's; the selected chip stays the only solid one
- [ ] `print` — printing `/cv` and a case study shows no toggle, no locale row and no strip, and the CV header carries its separator in the same place as on `main`
- [ ] `pdfs` — the seven committed PDFs are visually identical to `main`

| --- | --- | --- | --- |
| `once` | integration | ❌ | Assertable without a browser: no `ThemeToggle` import outside `src/app/`, and one occurrence in the built HTML of each route under `out/` |
| `corner` | manual-only | — | Inset and "reads as a corner" are pixel judgements across six pages and two widths |
| `themes` | e2e | ❌ | Click the control, assert `data-mantine-color-scheme` flips and persists across a navigation |
| `cv-phone` | e2e | ❌ | Deterministic, not a judgement: at a 360px viewport, assert the toggle's rect does not intersect the `h1`'s **text range** rect (its block box always spans the column) |
| `lift` | manual-only | — | Visual spacing judgement after a removed row |
| `article-nav` | manual-only | — | Whether the row reads right with one child is a look, not an assertion |
| `toolbar` | manual-only | — | A tint value is a design judgement; `/preview` in both schemes is the tool |
| `chip-hover` | manual-only | — | Contrast against two different grounds, judged by eye — and note a capture only shows it with the hover flag above |
| `print` | manual-only | — | Print preview; no print-rendering assertion exists in the suite |
| `pdfs` | integration | ❌ | `pnpm content:pdf --check` catches a *missing* re-render, not a visually wrong one. Comparing the decompressed glyph-show and rectangle operators between two PDFs does, and is what was run here |

Closes #40

https://claude.ai/code/session_01RW5ZEtg1ibnLPTGsXT6wn6

---


vzakharov:
Proposed squash title/body:

```
feat: #40 the theme toggle comes from the layout (pr #41)
```

```
Every page on the site rendered its own theme toggle -- two through a
CornerHeader wrapper, four in a Group of their own. An opt-in that every
consumer takes is not an opt-in: a page that forgot it shipped without a
toggle and nothing said so, and the hand-rolled sites were where
print-hidden had gone missing once already.

The toggle now lands once, from the app layer. RootLayout renders a
ThemeCorner that pins it to the page's top-right corner, anchored to the
page rather than to a header's box -- so it no longer tracks each page's
reading measure, and a page renders nothing for it. CornerHeader retires
with its last caller, and the case-study nav keeps only its back-link.
The CV's centred title is the one first line that reaches that corner,
and its header takes the clearance at phone width.

The CV's locale and .pdf controls move onto a tinted strip, which
replaces the header's hairline on screen and leaves it on paper. It is
the site's first filled surface, so the colour tokens gain a
surface/surface-strong pair -- two of them, because a chip hovering to
the same tint as the strip it sits on would be invisible. The second
settles a chip's hover, which had copied the current chip exactly:
hovering an unselected language painted it the same solid black as the
selected one, leaving the row with two chips both claiming to be
current. Solid foreground now means only that.

Previewing all of it corrected the preview skill in three places: its
documented capture floor is escapable through CDP viewport emulation,
which is what caught that CV collision; a hover rule never renders in a
capture at all, because headless Chromium reports (hover: none) and the
project's hover mixin takes its touch branch; and the theme attribute it
had the agent grep for was next-themes', not Mantine's. The seven
committed PDFs re-render with their ink unchanged.

Closes #40

```

---

vzakharov:
### Attestation

Verified `43b8ad7` (`claude/40-theme-toggle-in-layout-dkz48y`) against `main` @ `e42b636` "feat: refresh the CV's roles and controls, rework the theme picker (pr #36)" — the tip as of this check, and already contained in this branch, so the base merge was a no-op.

- CI-only buckets — **none exist: this project has no `/test-on-gh` lane, hydrated or stub**

Two caveats on what that record covers. The vet run's app-level check is `pnpm build`, which type-checks and statically renders every route but says nothing about how the change *looks* — and this PR is a visual change. The judgement rows of the QA checklist in the PR body (`corner`, `lift`, `article-nav`, `toolbar`, `chip-hover`, `print`) are what the earlier `/preview` session settled by eye; they remain unautomated. Note in particular that `chip-hover` is invisible to a plain headless capture, per the `/preview` correction this branch carries.

The working-artifact trees were swept before this head: `docs/issue/40/`, `docs/plans/` and `docs/remove-before-merging/` are all gone, and `docs/` contributes nothing to the diff against `main`.

---

PR #42, opened 2026-09-11
chore: sync the agent infrastructure — /plainly, /spinoff, a gated /finalize merge, and session-image extraction

vzakharov:
Pulls the agent infrastructure forward from the repo this one vendors it from, in two passes: watermark `4998149` → `0706dbb` → `a3ec24e`, 23 candidate commits triaged.

That repo renamed itself along the way — `vzakharov/agent-project-boilerplate` is now `vzakharov/muthur` — so the watermark's `repo` follows it, and so does the `adopted` key for the source's own copy of this skill, which is what the candidate log is filtered by.

Four skills are renamed or added, which is the operator-visible part: `/tighten-docs` → `/tend-prose`, `/sync-agent-boilerplate` → `/update-muthur`, and `/plainly` and `/spinoff` are new. The sync skill's watermark is `watermark.json`, the source's own filename.

## Triage — first pass (`4998149` → `0706dbb`)

| --- | --- | --- |
| 85f47c9 stop asking squash bodies for editing traps | take (mostly already had) | `dc31766` reached the same end state here independently; converged on the source's structure so the file stops drifting |
| 7959311 stage harness urgency behind the loop's own steps | translate (half) | Took the staging rule. Dropped the companion rule labelling loop changes `feat:`/`fix:` — the source's own commit says adopters invert it, and here the loop is infrastructure, so these stay `chore:` |
| f08b5fc keep the session link out of squash messages | take | The harness attribution block writes a `Claude-Session:` URL into every commit message, and the squash discards exactly the commits it is written for |
| 3051955 title a sync's squash record by what changed | take | Landed in this skill's Step 8; `upstream.json` reads `watermark.json` here |
| 722579b ship the sync path as an unhydrated stub | translate (rename only) | Took the rename (landing as `/update-muthur` once 8320b54 below is triaged). Rejected the stub-ification: the source has no source any more, this repo does, and its procedure is hydrated |
| d927329 absorb the built-in `/plan` collision | take | The collision is identical here — `plan` is a client-side built-in that never reaches the agent |
| 5f41dff run the skill-catalog check inside the vet run | translate | `scripts/vet.sh` is a local rewrite, so the check joined this repo's fan-out rather than the source's line |
| cd672c6 add `/spinoff` | offer → **adopted** (+ take the 40-line cap) | See "New skills" below. The body cap rode along and was taken |
| 44a1aa4 move the catalog off a path an adopter can own | translate | Repointed `check-skill-catalog.sh`, the skill's Step 4a and the `declined` key. The catalog itself stays declined and absent, so those assertions still skip |
| 3888707 split the repo resolver out of watch-tick-common.sh | take | Both callers are adopted verbatim; `scripts/lib/gh-repo.sh` is new |
| f526fd9 give the template-fork path a detemplating skill | offer → declined (+ take the `gh` notice) | See below. Took the session-start hook's gh-missing report, `/override-gh`'s litmus-test paragraph, and the environment-setup-script clause |
| aeb1062 write commit SHAs bare in GitHub comments | take | A backticked hash does not auto-link, which costs a review reply the one thing it was posted for |
| 61493a7 make the language decision a hydration item | translate | Written as this repo's answer (English) rather than the source's adopter stub, with the site's own `en`/`ru` content named as a separate decision |
| 29559e6 count characters, not bytes, in the squash width caps | take | `${#…}` counts bytes under `dash`, roughly halving the cap on Cyrillic |
| 80bbb29 label agent-authored comments in the export | take | `$GH_TOKEN` is the operator's identity, so the export's label is what tells an agent reply from operator guidance |
| 1a51f23 refer to the agent as singular they | take | One straggler here (`/preview`); the rest arrived with the files above |
| 0706dbb add a negation lens and rename the pass `/tend-prose` | take | The lens and the rename both hold here unchanged |

## Triage — second pass (`0706dbb` → `a3ec24e`)

| --- | --- | --- |
| 8f2b2cc gate an agent-driven PR merge on an uneventful finalize | translate | Step 8 and `/handle`'s non-forwarding clause land as written. One local clause added: here the merge is the deploy, so `and merge` authorizes a release |
| f4de0e1 state how an agent explains things | offer → **adopted** | `/plainly` plus the two files CLAUDE.md imports. See "New skills" below |
| 1482304 lead the README on what the repo is | skip (not adopted) | `README.md` is the source's description of itself; this repo's describes the site. Recorded in `declined` so it stops surfacing |
| c3b797e rename the repo to muthur | translate | `repo` and the `adopted` key follow; the source's term for itself ("boilerplate" → "template") follows in agent-facing prose only |
| — its `scripts/check-repo-identity.sh` | skip (not adopted) | It keys on the catalog this repo does not vendor, so it exits 0 here. The source's own `vet.sh` calls it the one line there that is that repo's alone |
| 8320b54 rename `/sync-muthur` to `/update-muthur` | take | The `adopted` key has to follow the source's path or the skill's commits stop surfacing — and the command follows too, on the source's own reasoning. See below |
| 463c59a let a session decide for itself whether a task needs a plan | take | `plan or go: <task>` and `plan-or-go.md`; every path it names exists here |
| a3ec24e pull operator-attached images out of the transcript | take | Hook, extractor and `scripts/lib/media.py` verbatim; `extension_for_bytes` moves out of the GitHub exporter so the hook does not depend on the issue loop |

## New skills offered

Each is recorded in the watermark, so the question does not come back — flip any of them by moving its entry between `adopted` and `declined`:

- **`/plainly`** — **adopted.** How an agent explains something to a person: cause first, in the reader's nouns, each step saying why the next one followed, an unknown cause read as unfinished investigation. Six named defects so a bad report can be called out in one word. `voice.md` and `operators.md` are imported by CLAUDE.md, ~3KB resident. `operators.md` lands verbatim rather than rewritten: the source's one entry is this repo's operator, the same person's own recorded preference.
- **`/spinoff`** — **adopted**, on the operator's call. Seeds a sibling repo out of the project you are standing in, carrying the foundation (layer boundaries, lint discipline, build and deploy shape, the agent loop) and leaving the product behind; the new repo's watermark points at the root, not at this one, so a chain never makes a sync walk an ancestry. This repo has already been spun off into another, so the operation is demonstrated here rather than hypothetical. Two translations, neither visible to the closure check: the relative link to the source's `catalog.md` resolves to nothing here; and the stackless-`vet.sh` rule it cites as `CLAUDE.md` § "Vetting" is the source's clause, declined here, so the skill states it itself. `lineage` joins the watermark's documented fields, since `/spinoff` writes it and cites that section as owning the contract — this watermark still carries none, having been filled in by hand.
- **`/detemplate`** — declined permanently. Converts a fresh "Use this template" fork into a project; this repo acquired its infrastructure by subset copy, long past the one moment the skill exists for.

## The names now travel

The source renamed its own copy to `/update-muthur` and argues the name should
travel, being the infrastructure's rather than each repo's. The first pass kept
`/sync-agent-infra` and the local watermark name `source.json`, on the grounds
that nothing here is called muthur. Review took both, and the argument was wrong
independently of that: `/update-muthur` is `npm update`'s verb — what gets
updated is the vendored copy in whatever tree the command runs in, never the repo
it came from, which the skill only ever reads. That reads the same at every link
in the chain, which is why the name travels rather than being re-coined per repo.

So the skill is **`/update-muthur`** at `.claude/skills/update-muthur/`, with the
watermark at **`watermark.json`** — the filename every downstream adopter has on
disk. Nothing locates either by path anyway: Step 1 finds a watermark by the
`repo` and `lastSyncedSha` it carries, which is what made both names free to
begin with, and what the first pass mistook for a reason to keep its own.

Two things fall out. `/spinoff` drops to two translations from four — with the
names identical on both sides, its § "The watermark" pointer and its paragraph on
what the target's sync skill is called both land verbatim. And the watermark's
note on the `adopted` list loses its worked example, since every adopted path is
now spelled the same in both trees; the note says that outright, a rule with no
live instance being the one that gets forgotten.

The source's own name for itself is re-spelled in agent-facing prose only. The
published case studies and the LinkedIn drafts name the repo as it was called when
they were written, and link to URLs GitHub still redirects.

## Watermark

`adopted` gained, across the two passes, seven paths that were vendored here but unlisted — so their source commits never reached a candidate set: `.claude/settings.json`, `.claude/hooks/`, `.claude/skills/plainly/`, `scripts/lib/gh-repo.sh`, `scripts/lib/media.py`, `scripts/extract-session-images.py`, `scripts/test_authorship.py` — plus `.claude/skills/spinoff/`, which moved out of `declined`. `README.md` and `scripts/check-repo-identity.sh` joined it.

## Verification

`./scripts/vet.sh` green at 18bc1f1. The extractor was run against this session's own transcript (1023 records read, no attachments, exit 0), the hook is silent on a transcript it cannot find, and the exporter's 12 tests still pass.

https://claude.ai/code/session_01NaEVxBThGy9pgVBiHd59tC


.claude/skills/update-muthur/SKILL.md
vzakharov, line 1: sync-muthur

.claude/skills/update-muthur/watermark.json
vzakharov, line 1: watermark.json

.claude/skills/update-muthur/SKILL.md
vzakharov: Done — 57de89d. Kept your verb rather than the source's: it renamed `/sync-muthur` to `/update-muthur` because the command is named for the tree it runs in, which there is muthur itself. One link down that inverts — run here, `/update-muthur` reads as updating muthur. So the repo travels into the name and `sync` stays.

---

.claude/skills/update-muthur/watermark.json
vzakharov: Done — 57de89d. This one takes the source's spelling outright: `watermark.json` is what every downstream adopter has on disk, and `/spinoff` already carries a list of the path shapes a caller may show up with rather than looking one up, so the local name only cost a reader of two trees a translation.

---

vzakharov:
Proposed squash title/body:

```
chore: add /plainly and /spinoff, rename the prose and sync passes (pr #42)
```

```
The agent infrastructure here is vendored from vzakharov/muthur, which
has kept editing it. This pulls that forward across two windows, 23
source commits triaged one at a time and ported by intent rather than
by patch.

Two skills are new. /plainly states how an agent explains something to
a person — cause first, in the reader's nouns, an unknown cause read as
unfinished investigation rather than as a style constraint — and
CLAUDE.md imports two of its files, so the rule governs every reply
whether or not the skill is invoked. /spinoff goes the other way: it
seeds a sibling repo out of this one, carrying the foundation and
leaving the product behind, and points the new repo's watermark at the
root so a later sync never walks an ancestry. The watermark's contract
grows a `lineage` field to carry that.

Three new entries reach the loop. `plan or go: <task>` is a conditional
go-ahead that hands the agent the judgment of whether the work needs a
plan at all, scoped to the one message it appears in. `/finalize and
merge` lets the agent press the merge button, but only when the run
turned up nothing to decide — here that authorizes a release, since
merging to main is the deploy. And a UserPromptSubmit hook writes
operator-attached images out of the transcript into gitignored tmp/,
where they can be looked at and, if worth keeping, moved into the repo
deliberately.

Two passes are renamed. /tighten-docs is /tend-prose, which also gains
a lens for prose surviving only to deny what the change removed; and
the sync pass takes the source's own name, /update-muthur, with its
watermark at watermark.json. That watermark's `adopted` list holds the
source's paths because it is what filters the candidate log — an entry
left behind when the source renames silently drops every future commit
under the new path.

```

---

vzakharov:
### Attestation

Verified 18bc1f1 (`claude/sync-agent-infra-plan-w1al79`) against `main` @ 5a5bd5e "feat: #40 the theme toggle comes from the layout (pr #41)".

- CI-only buckets — **`/test-on-gh` is declined here, not a stub: nothing in the suite needs credentials, a service or a browser**

`/check-merge` reports `contained` — `main` has not moved since, as of this check.

---

.claude/skills/update-muthur/SKILL.md
vzakharov: Corrected to `/update-muthur` — 18bc1f1. My argument for the other spelling was wrong independently of the typo: I read `update-muthur` as naming what gets updated *to*, when `npm update lodash` updates your lodash, not lodash's repo. The verb reads the same at every link, so the name travels, which is what the source argued in the first place.

That also puts `/spinoff` back to the source's wording on what the target's sync skill is called — two translations there now instead of four.

---

PR #43, opened 2026-09-14
feat: latestageagentic.com, built from this repo, and its dictations

vzakharov:
Two spoken passes at a new project, written down so they stop living in audio files; the pipeline that will do it for the rest of them, after eight rounds of review; and the channel's opening post, written as text rather than spoken.

**The site it will be served from, built out of this repository.** `latestageagentic.com` is a second Next app rather than a second repository: `apps/vova/` and `apps/lsa/` each hold a router, a `public/` and a config, and both build from the one `src/`. That works because `next build` needs nothing else in a project directory — no `package.json`, no workspace, one lockfile — and because FSD already forbids two page slices from reaching each other sideways, which is exactly the relationship two sites' pages have. So `steiger`, the `boundaries` rules and the `@/*` alias are untouched.

Each app names itself in its `next.config.ts`, and `src/shared/config` holds both sites under one `SiteConfig` type and refuses to load without a name — so a build that forgot which site it was would fail rather than publish one site's copy under the other's domain. It is a lookup rather than a zod parse: client components reach that module, where the schema would ship in the chunk.

**Every build is entered in its app directory**, which is what makes the layout work at all: next-intl's plugin checks its config path against the working directory and hands Turbopack the same string to resolve against the project, and only one directory satisfies both. `PUBLIC_DIR` therefore keeps resolving off the working directory as it always has, and the render scripts are entered there too. Root `pages/` goes away with it — Next looks for a Pages Router inside the project directory only, which is now a level below `src/pages/`.

**The deploy grows a second lane.** A repository gets one Pages site, so `vovazakharov.com` keeps going out through this one's and `latestageagentic.com` is force-pushed by `scripts/publish-lsa.sh` to the `gh-pages` branch of `vzakharov/latestageagentic.com` — a repository with no source, whose Pages is set to deploy from a branch, so the push is the deploy and nothing runs over there. The script refuses an `out/` missing `CNAME` or `.nojekyll`, both of which fail silently at the far end: a dropped custom domain, and a `_next/` that Jekyll strips. The gate now reads the commit scope as a site name — `feat(lsa):` publishes one, `feat(vova):` the other, anything else both. A manual run has no subject to read a scope off, so `workflow_dispatch` takes a **site** picker instead, defaulting to both; naming one is what lets an unmerged branch publish that site alone, without deploying the other from a tree nobody has merged.

[...]

PR #44, opened 2026-09-14
docs: plan the late-stage-agentic spinoff

vzakharov:
## Summary

- **Plans the move of `writing/late-stage-agentic/` into its own repository**, which `writing/late-stage-agentic/plan.md` has always said would happen and PR #43 is the point where there is enough of it to happen. One file: `docs/plans/spinoff.draft.do-not-implement.md`.
- **`@.claude/skills/spinoff/SKILL.md` owns the procedure and is not restated.** What a run of it cannot derive on its own is what the plan holds: the settled Step 1 answers (`vzakharov/latestageagentic.com`, public, the same stack), the per-path triage, and the sequencing.
- **The triage is the bulk of it**, sorted the way Step 4 lands paths — copies bound for the target's `main`, rewrites bound for its PR #1 — with the calls the criterion does not settle alone argued individually: the content-pipeline rule (product-scoped on its face, and the single most valuable thing travelling, because a wiki is almost entirely that pipeline), the session-start hook that splits into a travelling proxy shim and a stack-bound install, the deploy workflow whose domain must not travel, and the lockfile that is regenerated rather than copied.
- **Read from this repo's side, the PR is an amendment to `/spinoff`.** Two gaps are visible before the run and the plan commits to closing both. The skill never says **when** to run relative to the material — Step 1 reads the caller at HEAD and never asks what HEAD holds — so content still under review travels through the product hatch and every later review outcome is applied twice, in two repos. And it states that a new repo's `main` carries a stub `vet.sh` without saying what a caller whose own `vet.sh` is real should write there: this repo has no stub to copy, and copying the real one reds the new repo from its first commit.
- **The material leaves; the tooling stays.** `writing/late-stage-agentic/` goes with one tombstone for the whole retirement. The four dictation skills and `scripts/transcribe.py` are tooling rather than material and stay here as well as travelling — two copies with no path between them, since both repos sync from `vzakharov/muthur` and it carries none of the four. That cost is accepted, and the plan's DRY notes say why the two alternatives are worse.
- **Sequenced after PR #43 merges.** Running today off its branch is procedurally fine — the skill permits it — and wrong here: findings are still open, and that is exactly the case the first amendment above exists to name.

## QA Checklist

- [ ] `amendments` — the two gaps in `/spinoff` are the PR's headline. Read them against the skill and say whether they are real, and whether either belongs in `CLAUDE.md` rather than the skill.
- [ ] `timing` — the skill permits running today; the plan waits anyway, for a reason particular to this branch. Confirm the trade: one merge against applying #43's remaining review outcomes twice.
- [ ] `two-copies` — the dictation set living in both repos has no reconciliation path at all. Check that the DRY notes state the cost the way you read it, and that "revisit if they actually diverge" is the right disposition.
- [ ] `triage` — read the three triage tables against your own sense of the tree. The rows to argue with are `.claude/rules/content.md` and `.claude/rules/writing.md` (rewrites, not copies), `scripts/lib/media.py` (stays — shared with the loop's image extraction), and `scripts/check-notes-length.sh` (stays).
- [ ] `hatch` — the product hatch is honored only when named in the invocation, and it is the one thing that genuinely leaves. Check that what it carries is what you expect in the new repo: the two dictations, the two drafts, `plan.md`, and the recordings under `docs/remove-before-merging/`.
- [ ] `nothing-ran` — `git diff --stat origin/main..HEAD` touches `docs/plans/` and `docs/remove-before-merging/` and nothing else. No spinoff has been run and no repository has been created.

| --- | --- | --- | --- |
| `amendments` | No | n/a | Whether a skill is missing something is read, not measured. |
| `timing` | No | n/a | Depends on when #43 lands, which is the operator's. |
| `two-copies` | No | n/a | An accepted cost, and accepting it is theirs. |
| `triage` | Partly | No | `check-skill-catalog.sh` catches a reference left dangling by the move, but only once the move has happened. Nothing can check a triage before its run. |
| `hatch` | No | n/a | What counts as this project's material is the operator's definition. |
| `nothing-ran` | Yes | Yes | `git diff --stat` is the check. |

Draft, and staying draft until the plan is approved and the spinoff has run.

https://claude.ai/code/session_01Hhsu3uVmJyAJquBFxDr5kq


vzakharov:
Proposed squash title/body:

```
chore: close the two gaps /spinoff's first run found (pr #44)
```

```
/spinoff ran for the first time here, and two things it does not say
showed up in the run. It reads the caller at HEAD and never asks what
HEAD holds, so material still under review travels through the product
hatch and every later review outcome gets applied twice, in two repos.
And it requires a new repo's main to carry a vet.sh stub that exits 0
naming no stack, while saying nothing about what a caller whose own
vet.sh is real should write there — this repo had no stub to copy, and
copying the real one reds a new repo from its first commit. Both are
now in the skill.

writing/late-stage-agentic/ leaves for vzakharov/latestageagentic.com,
where it becomes a site and a channel of its own. One tombstone stands
in for the whole retirement, and it records what no file can enforce:
deleting claude/late-stage-agentic-phnz8v destroys the recordings
behind every piece published from them, because
docs/remove-before-merging/ is swept at merge and the branch is the
only thing keeping them reachable.

The dictation set stays — /dictation, /afterword, /subtitles,
/dictation-to-post and scripts/transcribe.py — so the skills outlive
the material they were built for and a recording made here later has
them waiting. What that costs is two copies with no path between them:
both repos sync from vzakharov/muthur, which carries none of the four,
so a fix made in one is a fix made in one.

```

---

vzakharov:
Superseded by https://github.com/vzakharov/vovazakharov.com/pull/45, which keeps the project in this repository and builds it as a second site instead of moving it out. This one is ready to close; the branch stays either way, so the triage in its plan file remains readable.

---

PR #45, opened 2026-09-15
docs: plan latestageagentic.com as a second site in this repo

vzakharov:
## Summary

- **Plans `latestageagentic.com` as a second site built from this repository**, not as a repository of its own. One file: `docs/plans/two-sites-one-repo.draft.do-not-implement.md`. This replaces the spinoff approach on #44, which assumed the project would move out; nothing moves, so `/spinoff` is not run and #43 lands unchanged.
- **The layout is `apps/vova/` + `apps/lsa/` over one shared `src/`.** Each app holds only routing, `public/`, `next.config.ts` and a `tsconfig.json` that reaches back to the root `src/`. Both sites' page slices stay in one `src/pages/` — FSD slices may not import each other sideways, which is exactly the relationship two sites have — so `steiger`, the `boundaries` patterns and the `@/*` alias are untouched by the whole change.
- **A spike settled the load-bearing question before the plan was written.** `next build <dir>` against Next 16.0.3 in this tree needs no `package.json` in that directory: `@/` resolves across roots, `.next/` and `out/` land inside the app dir, the repo root stays clean. So there is **no pnpm workspace, no Turborepo, one lockfile** — the alternative would have forced both apps to declare every dependency separately.
- **CI builds both and pushes one.** `vovazakharov.com` keeps deploying to its own Pages exactly as today; `apps/lsa/out` is force-pushed as a single orphan commit to `vzakharov/latestageagentic.com`, whose Pages source is a branch — so the receiver carries no workflow and both builds stay in one pinned toolchain. That is the reason for preferring it over Cloudflare, which would put the second build on a runner no agent here can reach.
- **The gate learns to read the commit scope.** `feat(lsa):` publishes LSA only, `feat(vova):` vova only, an unscoped `feat:`/`fix:` publishes both. The scope is already in the gate's regex today and thrown away.
- **The by-hand half is written out as a runbook**, because none of it has an API this repo can reach: creating the receiver, generating the deploy key and installing its two halves, the eight DNS records for the apex, pointing Pages at the published branch, and forcing HTTPS once the certificate lands.

Four open questions are numbered at the end of the plan, each with a recommendation already in force in the text — so silence resolves them.

## QA Checklist

- [ ] `shape` — read the `apps/` + one-`src/` layout against your own sense of the tree. The row to argue with is both sites' page slices sharing `src/pages/`.
- [ ] `spike` — the no-workspace decision rests entirely on one verified build. Check that what it proved is what the plan claims it proved.
- [ ] `config` — `SITE_CONFIG` selected by a zod-parsed `NEXT_PUBLIC_SITE` pinned in each app's `next.config.ts`, versus the per-app module alias the DRY notes rule out. This is the one thing that touches every call site if it is wrong.
- [ ] `gate` — `feat(lsa):` / `feat(vova):` / unscoped. Confirm that a `feat(cv):` publishing both sites is the behaviour you want from an unknown scope.
- [ ] `key` — one deploy key, private half a secret here, orphan force-push to the receiver. Check the posture: no third-party action holds the key, and rollback is a workflow dispatch from an older ref rather than a revert on the receiver.
- [ ] `runbook` — the one section written for a person rather than an agent. Read it as the person who will run it: is any step guessable-wrong?
- [ ] `day-one` — question 1. What the LSA site serves before any content moves.
- [ ] `nothing-ran` — `git diff --stat origin/main..HEAD` touches one file under `docs/plans/`. No repository was created, no key generated, no source moved.

| --- | --- | --- | --- |
| `shape` | No | n/a | An architecture call, read rather than measured. Once implemented, `steiger` and `boundaries` hold it. |
| `spike` | Yes | Yes | It was run; the plan states what it showed. Re-runnable in `tmp/`. |
| `config` | Partly | No | The zod parse makes a wrong value fail the build, but nothing checks that the two configs are the right shape until the type exists. |
| `gate` | No | n/a | What an unknown scope should do is a convention, not a fact. |
| `key` | No | n/a | Secret handling; nothing in the repo can verify a key it must not read. |
| `runbook` | No | n/a | Followable-by-a-human is not a property a script can assert. |
| `day-one` | No | n/a | Product decision. |
| `nothing-ran` | Yes | Yes | `git diff --stat` is the check. |

Draft, and staying draft until the plan is approved.

https://claude.ai/code/session_01Hhsu3uVmJyAJquBFxDr5kq


vzakharov:
Proposed squash title/body:

```
feat: build latestageagentic.com from this repo as a second site (pr #45)
```

```
The project wanted a domain of its own, and the first answer was a
repository of its own. It does not need one: what a second site needs
is its own route tree, its own public/ and its own config, and all
three fit here beside the first.

apps/vova/ and apps/lsa/ hold exactly that, over one shared src/ —
including both sites' page slices, because FSD slices already may not
reach each other sideways, which is the relationship two sites have.
There is no workspace and no task runner: next build <dir> resolves
from the single root node_modules, so an app directory needs only a
config, a route tree and a tsconfig pointing back at src/. Which site
is being built is a build-time constant pinned in each app's config
and parsed on the way in, so an unset value fails the build rather
than serving the wrong site.

CI builds both. vovazakharov.com publishes to its own Pages as
before; the latestageagentic.com build is force-pushed as a single
orphan commit to vzakharov/latestageagentic.com, whose Pages serves a
branch and therefore carries no workflow of its own — which keeps one
pinned toolchain behind both sites instead of handing the second to a
runner nothing here can reach. Rollback is a dispatch from an older
ref, so the receiver never grows a history. CNAME and .nojekyll ride
inside the build, and the publish step refuses an out/ missing
either: a dropped custom domain and a Jekyll-stripped _next/ are both
silent at the far end.

The deploy gate now reads the commit scope it used to discard.
feat(lsa): publishes one site, feat(vova): the other, and an unscoped
feat: or fix: both.

```

---

vzakharov:
Closing: the plan belongs on the branch that carries the work, not on one of its own. It now rides `claude/late-stage-agentic-phnz8v` (#43) as 6effcf2, with the two questions this split had raised — when to land relative to #43, and whether to close #44 — collapsed out, since the answer to both is here.

---

PR #46, opened 2026-09-15
feat: sync the agent loop forward — /task, /polish, operator voice

vzakharov:
## Summary

- **The entry ladder routes on one question** — does the opening prompt ask for a change to this codebase? It does, and the work goes to the new `/task`, which makes the plan-or-not call itself; it does not, and the answer is the deliverable. This replaces the `plan or go:` opt-in, and `plan-or-go.md` retires with it.
- **`/issue` splits into transport and decision.** `/take-issue` exports and commits a thread and hands the number back, deciding nothing; `/issue` stays as a redirect. A new `.claude/hooks/prompt-issue-export.sh` does that export on `UserPromptSubmit` when a session's opening prompt ends in `#<N>`, so the thread is on disk before the agent reads the prompt.
- **The carve moves to `.claude/skills/plan/carving.md`**, loaded by `/plan` when work is beyond one PR and by `/go` to file what the plan proposed. Nothing is created at plan time — the plan file is already the gate.
- **`/polish` is the new home of `/dry` + `/tend-prose`**, with the scope resolution neither pass could do alone. `/go` Step 3 loads it, `/finalize` runs it ahead of its numbered steps, and the operator invokes it over work that took neither route. `polish:` joins the commit-type list as a branch-local type.
- **The session knows who it is talking to.** Startup is three hooks now — `gh-shim.sh`, `install-deps.sh`, `operator-voice.sh` — and the voice rule moves out of `/plainly` to `.claude/voice/`, with one entry file per handle. `check-skill-catalog.sh` widens from skill pointers to every `@`-reference into `.claude/` so CLAUDE.md's own imports are covered.
- **PR and issue exports stop growing without bound.** Review threads quote the reviewer's own line selection rather than GitHub's padding, the review section opens with a per-thread index, and past 400 rendered lines the bodies hoist into `threads/` and `comments.md`.
- **The vet fan-out drops to twelve, and the vendored tests come out of the tree with it.** The source's `catalog.md` dispositions `scripts/test_*.py` `never` — they cover its own scripts, `scripts/gh_export/` among them — so both files leave and join `declined`, `test_authorship.py` included, which had sat in the fan-out since pr #42. `scripts/gh_export/` itself stays adopted: it is what this repo runs. Why that disposition never reached the triage is filed upstream as vzakharov/muthur#82.

Seven source commits triaged, six ported; the triage table with every verdict is below. `watermark.json` advances to `61715d0`.

## Triage

| --- | --- | --- |
| `d21aa0e` a comment body is text, not a path | take | The GitHub-comments section is identical here, and `-f body=@<file>` misfires the same way |
| `3c3ce16` name and greet the session's operator | translate | Taken whole; `install-deps.sh` ships upstream as a TODO stub and carries this project's `pnpm install --frozen-lockfile` instead |
| `2e673f4` make /handle dispatch to /go imperatively | take | Same wording, same failure mode |
| `1b24fe7` route the entry ladder, move the carve | translate | `carving.md` states the filing inline against `gh` and drops the dedupe pass, `/propose-issue` being declined here; the `/issue` redirect's no-number branch names `/task` for the same reason |
| `3af9da4` export the issue a prompt ends in | take | Hooks and `lib.sh` land verbatim |
| `5049066` split long exports into an index | translate | The exporter lands verbatim. Its tests do not: `catalog.md` dispositions `scripts/test_*.py` `never`, and `check-muthur.sh` — new in this same commit — is the source's one vet line for everything measuring code an adopter vendors and never edits. Both are declined here and the two files are out of the tree (thread on `scripts/vet.sh`) |
| `61715d0` extract the quality passes into /polish | take | Both passes and all three call sites are adopted here |

**Also declined**, recorded in `watermark.json` so the question does not come back: `scripts/check-muthur.sh`, for the reason `check-repo-identity.sh` already sits there — but this repo now takes the decision that script encodes rather than only the script's absence. Every other `declined` entry was re-read and its condition still holds — `/propose-issue` in particular, this backlog still being one an operator reads in a sitting.

## QA Checklist

- [ ] `vet` — `./scripts/vet.sh` is green (all twelve fan-out checks plus the build).
- [ ] `operator-voice` — `bash .claude/hooks/operator-voice.sh` names the operator and prints their entry (verified: resolves @vzakharov from the session token).
- [ ] `issue-export-hook` — a payload whose prompt ends in a real `#<N>` writes the export and emits context; one ending in nothing, or in a number that will not fetch, exits 0 without failing the turn (both verified).
- [ ] `exporter` — `python3 scripts/export-github-item.py <n>` writes a PR export whose review section opens with a thread index (verified against PR #42, and again against this PR).
- [ ] `closure` — `bash scripts/check-skill-catalog.sh` passes, including CLAUDE.md's `@.claude/voice/voice.md` import.
- [ ] `settings` — a fresh session starts all three SessionStart hooks and all three UserPromptSubmit hooks without error.

| `vet` | yes | yes | `scripts/vet.sh` |
| `operator-voice` | yes | no | needs a live `gh` token; run by hand |
| `issue-export-hook` | yes | no | no test file yet; a candidate if the hook grows |
| `exporter` | yes | at the source | the suite lives in `vzakharov/muthur`, where `scripts/gh_export/` is edited |
| `closure` | yes | yes | `scripts/check-skill-catalog.sh`, in the fan-out |
| `settings` | no | no | observable only on a real session start |

https://claude.ai/code/session_01AJqCxTBA5xVCwEGghxWUan


vzakharov:
Proposed squash title/body:

```
chore: rework the loop's entry and quality passes, name the operator (pr #46)
```

```
The repo this one vendors its agent loop from moved, and four of its
changes are structural here. A new session's opening prompt now routes
on one question — does it ask for a change to this codebase? — rather
than defaulting to /plan, and the work goes to /task, which judges for
itself whether a plan is needed and whether that plan has to reach the
operator. The `plan or go:` opt-in it replaces retires with its page.

/issue splits along the seam it always had. /take-issue exports a
thread onto the branch and decides nothing; /issue stays as a redirect;
and a UserPromptSubmit hook does that export before the agent reads a
prompt ending in `#<N>`, so a session launched at an issue starts
holding it. The carve that /issue used to own moves to plan/carving.md,
loaded only by a plan that needs it — and, filing nothing until the
go-ahead flips the plan file, it makes untracked work splittable for
the first time.

The two quality passes get a home of their own in /polish, which
resolves the scope neither /dry nor /tend-prose could resolve alone:
both bottom out at unpushed work, so both read nothing on a branch
someone else wrote. /go and /finalize call it, the operator calls it
over work that reached neither, and a branch-local `polish:` commit
marks how far the last run got. Sessions also know who they are talking
to: startup is three hooks now, and the voice rule they serve moves out
of /plainly to .claude/voice/, one entry file per handle.

Underneath, PR and issue exports stop growing without bound — threads
quote the reviewer's own line selection, an index opens the review
section, and long exports hoist their bodies into sibling files. The
reference check widens from skill pointers to every @-reference into
.claude/, which is what covers CLAUDE.md's own imports. The vet fan-out
drops to twelve and scripts/test_*.py leaves the tree, the source's
catalog dispositioning those tests never: they cover scripts/gh_export/,
which is vendored byte for byte and edited only there.

```

---

scripts/vet.sh
```diff
@@ -70,7 +70,8 @@ scripts/run-parallel.sh \
   squash='scripts/check-squash-message.sh' \
   notes='scripts/check-notes-length.sh' \
   skills='scripts/check-skill-catalog.sh' \
-  export='cd scripts && python3 test_authorship.py' || status=1
+  export-authorship='cd scripts && python3 test_authorship.py' \
+  export-split='cd scripts && python3 test_export_split.py' || status=1
```
vzakharov: aren't these muthur-specific tests that are not supposed to run for adopters?

vzakharov: You're right, and the source says so in its own header. `check-muthur.sh` opens with the membership test — *"would the adopting repo run this?", not "is it a test"* — and puts in the bucket whatever "measures code an adopter vendors and never edits". `scripts/gh_export/`, `export-github-item.py` and both `test_*.py` are byte-identical to `vzakharov/muthur` here, so they're squarely in it.

What I got backwards was reading that script's `exit 0` downstream as a malfunction rather than as its answer. The watermark's decline said it "would certify nothing here", and I named the two tests individually in the fan-out instead. The skip *is* the decision: an adopter runs none of it.

Both lines are out in 7a1d0d5 — including `export=`/`test_authorship.py`, which had been there since pr #42 under the same misreading, so the fan-out goes to twelve rather than back to thirteen. The files stay adopted and carried; they're the exporter's contract, held where the exporter is edited. The watermark's `check-muthur.sh` reason is rewritten to say that instead of the opposite, and CLAUDE.md § "Vetting" now states the membership test for the whole list.

Logged in `writing/notes/the-five-percent.md` under "What it was handed, it treats as fixed" (×10): `test_authorship.py` being there already made "add its sibling" the only question the sync asked.

---

vzakharov: Second half of the same answer, having gone and read `catalog.md`: it dispositions `scripts/test_*.py` **`never`**, so those files should not be in this tree at all — not merely out of the fan-out. Both are deleted and `declined` in e6503cb. `scripts/gh_export/` stays `adopt`, being what this repo actually runs.

The reason the row never reached me is worth fixing upstream, so it's filed as vzakharov/muthur#82: `/update-muthur` reads the catalog only when a commit adds a **skill** (Step 4a, and `catalog.md` § "Two audiences" says the same), so a `never` on a script is invisible to a sync. The tests arrive as a side effect of porting the exporter's commits, which is the one path nothing checks.

---

vzakharov:
### Attestation

Verified a87c736 (`claude/update-muthur-k80c10`) against `main` @ 7e8c401 "docs: bring the README up to the two-site layout (pr #48)".

- `./scripts/vet.sh` — **green**, on this exact head (all twelve fan-out checks, plus both builds and the styles codegen)
- CI-only buckets — **/test-on-gh is declined here, not an unhydrated stub**: the watermark records why, and nothing in the fan-out is too slow or too heavy for a local run

Two base merges are folded in since the last push. 7499592 (pr #43) conflicted in `CLAUDE.md` and `writing/notes/the-five-percent.md` and was resolved by hand — the vetting roster takes main's two-site wording with this branch's count of twelve, the stack-moves-with-the-operator paragraph takes main's two-item version with this branch's hook filenames, and the notes file takes main's squeeze of the four shared accounts with this branch's own entry re-added on top, paid for by tightening five others. 7e8c401 merged clean and is independent of this diff: a README rewrite and one error-message string in `scripts/render-pdf.ts`, neither of which this branch touches.

---

Issue #47, opened 2026-09-15
CV PDFs link to a dev server, and a re-render churns every PDF


vzakharov:
Two problems in the printed-PDF pipeline, found while finalizing #43 by diffing a
re-render against the committed file. Both predate that PR — `main` carried them
before it and carries them now — and they are independent of each other, so they
can land separately. They are filed together because the same 16 differing bytes
per file surfaced both.

## 1. The CV PDFs link to a dev server (bug)

All four CV PDFs carry two link annotations pointing at the machine that rendered
them:

```
/URI (http://localhost:41593/)
/URI (http://localhost:41593/case-studies/playgram)
```

The port is whatever the render's dev server happened to get, so the links are
dead for every reader. The case studies' own PDFs are clean — this is the CV's
alone.

The cause is that the printed link's **text** is absolutized and its **href** is
not. `src/pages/cv/ui/case-study-link.tsx` renders the visible address as
`` `${cv.website}${href}` `` — `vovazakharov.com/case-studies/playgram`, which
reads correctly on paper — while passing the same relative `href` to the anchor,
which Chromium then resolves against the page origin. `src/pages/cv/ui/cv-sheet.tsx`
has the same shape at its two `<InternalLink href="/">` call sites.

`printedUrl()` in `src/shared/config/site-config.ts` is exactly the helper for
this: it returns the absolute `href` alongside the scheme-less text, and its
docstring already states why the two differ. It is used by
`shared/content/plugins/rehype-media-embeds.ts` and
`src/pages/case-studies/ui/printed-from.tsx`, but the CV never reaches it. Routing
the three print-path call sites through it is the fix; the on-screen anchors
should stay relative.

**Verify with** `strings apps/vova/public/cv/cto/en.pdf | grep -o '/URI ([^)]*)'`
after re-rendering — no `localhost` should survive, and the visible text should be
unchanged.

## 2. A re-render rewrites every PDF for nothing (hygiene)

`pnpm content:pdf --check` compares each PDF's **source set** against
`pdf-renders.json`, so any edit under `src/shared/config` or `src/shared/ui` — a
new export, a new component — re-flags all seven renders even when nothing on a
printed page moves. That part is the design working: the manifest deliberately
hashes inputs rather than output.

What is avoidable is the churn. Re-rendering after such an edit produces pages
that are byte-identical apart from `/CreationDate` and `/ModDate` (plus the dead
links above, which carry a fresh port). So the honest fix — re-record the hashes —
[...]

PR #49, opened 2026-09-16
fix: #47 print the CV's links absolute, and keep unchanged PDF bytes

vzakharov:
## Summary

- **The CV's printed links pointed at the machine that rendered them.** All four
  committed CV PDFs carried `http://localhost:41593/` and
  `http://localhost:41593/case-studies/playgram` — the printed link's _text_ was
  absolutized while its _href_ stayed relative, so Chromium resolved it against
  the dev server. Both call sites now take `printedUrl()`, the helper that
  already returns the absolute href alongside the scheme-less text and already
  serves `rehype-media-embeds` and `printed-from`. Every link annotation keeps
  its exact `/Rect`, so the printed page is unchanged and only the targets moved.
- **`InternalLink` renders a link once per medium, and screen routing is
  untouched.** It renders a `print-hidden` anchor on the relative href
  `next/link` needs for a client-side route and a print-only `Anchor` on the
  absolute one paper needs; a call site names the address once and says only
  whether paper also spells it out. Every internal link gets the pair, because
  every one carries that relative href and any page can be printed — an external
  link needs none, being absolute already. `InternalButton` is the one
  exception, a button being something to press. All four PDFs re-render
  byte-identical across the change, and all seven come back `kept`, so no
  printed page moved. `pick()` joins `shared/lib/collections` beside
  `class-names` — `vova/no-redundant-property-copy` names it as the remedy and
  nothing here provided one.
- **A re-render no longer rewrites a PDF for nothing.** `render-pdf.ts` compares
  a fresh render against the committed file with the two things the renderer
  never settles discounted — `/CreationDate` and `/ModDate`, and the names
  Chromium gives a tagged PDF's structure-tree nodes — and keeps the committed
  bytes on a match. Since staleness is decided by hashing sources, any edit under
  `src/shared/config` or `src/shared/ui` re-flags all seven PDFs; the run's
  output is now always safe to commit rather than needing a byte diff read by
  hand to decide whether to keep the old files.
- **`cv.website` is gone from the message catalogue** (`/dry`'s find). Routing
  the case-study line through `printedUrl()` left it with one consumer, and it
  spelled `vovazakharov.com` a third time beside `SITE_CONFIG.url` — three homes
  for one domain, none of which would catch a missed edit.

### On the report that the links work (issue comment)

Seven of the nine link annotations were always fine: email and website on page
1, and GitHub, LinkedIn, X plus the email and website pair in the contact block.
The two broken ones sat where a reader is least likely to test them — the
[...]

PR #54, opened 2026-09-16
chore: sync the agent loop forward — a PR export per turn, colocated skill pages

vzakharov:
## Summary

- **`/handle` re-exports its branch's PR before every turn.** A `UserPromptSubmit` hook resolves the prompt's target to a PR through `gh` and re-exports it; the turn then commits what it read, so the next export arrives as a `git diff` against it. An empty diff is the answer to "has anything changed?", not the absence of one — and the committed sequence shows a reviewer which state of the PR each turn was answering. A bare `/handle` now targets the branch the session is on.
- **A session's opening prompt gets routed before it gets worked.** The entry ladder sits in CLAUDE.md, resident and routinely skipped, so a change asked for in prose got made directly — no plan-or-not call, no quality passes, no PR. `.claude/hooks/prompt-route-notice.sh` puts the call at the moment it is made, pointing at the ladder rather than copying its rows.
- **Three minority-entry skill sections move to colocated pages**, each behind a resident pointer naming its trigger: `/plan`'s native plan-mode recovery, `/squash-message`'s working-file selection, `/take-issue`'s video-frame extraction. `check-skill-catalog.sh` gains assertion 5 — the reverse of assertion 1, since a page nothing points at fails as silently as a pointer to nothing.
- **An export is one file however long the PR.** Thread and comment bodies no longer hoist into `threads/` and `comments.md`; the index survives, being how a reader picks threads to open. `gh_export/split.py` becomes `gh_export/index.py`.
- **`/pr` refreshes over REST.** `gh pr edit` asks for project cards on every edit, so where classic projects are deprecated it fails and leaves the PR exactly as it was.

The watermark advances to the source's `69c1908`. Four source commits were triaged; the two declined paths (`update-muthur/catalog.md`, `scripts/test_*.py`) stay declined for the reasons already recorded.

## QA Checklist

- [ ] `route-notice` — open a fresh session with a prose prompt that asks for a change. The turn's context should carry the routing notice; a prompt starting with `/`, and every prompt after the first, should not.
- [ ] `pr-export` — say `/handle <a branch with an open PR>` in a fresh session. The hook should report `docs/pr/<n>/pr.md` exported fresh before the turn. Say it again after a comment lands on that PR: the report should carry the arriving comment as a `diff` block.
- [ ] `pr-export-quiet` — say `/handle` on `main`, and `/handle` on a branch with no PR. Both should be silent.
- [ ] `export-layout` — `python3 scripts/export-github-item.py 49`. One `pr.md`, no `threads/` and no `comments.md`; every index row's `→ [↓](#tNN)` should resolve to an anchor further down the same file.
- [ ] `catalog-check` — `bash scripts/check-skill-catalog.sh`. Assertion 5 should list as run; delete a pointer to one of the three new pages and confirm it fails.
- [ ] `pr-refresh` — run `/pr` on a branch that already has a PR and confirm the body updates (this is the path `gh pr edit` was failing on).
- [ ] `vet` — `./scripts/vet.sh` green.

| `route-notice` | yes, partly | no | The hook's three branches are testable by piping a payload; whether the agent then routes is not. Smoke-tested by hand this session. |
| `pr-export` | yes, partly | no | Same shape: payload in, context JSON out. Smoke-tested against PR #49 this session. |
| `pr-export-quiet` | yes | no | Pure predicate — a `scripts/test_*.py` case if this repo ever ran one, which it declines to. |
| `export-layout` | yes | no | The source's own `test_export_render.py` covers it, and is declined here — the suite stays where the code is edited. Verified by hand against PR #49. |
| `catalog-check` | yes | **yes** | `scripts/vet.sh` runs it. |
| `pr-refresh` | no | no | Needs a live PR and a GitHub token; the failure it fixes is environment-specific. |
| `vet` | yes | **yes** | Is the suite. |

https://claude.ai/code/session_01EL1vFft5BczURQUXT7hXek


vzakharov:
Proposed squash title/body:

```
chore: a PR export per turn, and the entry ladder at the prompt (pr #54)
```

```
Two ends of the agent loop answered from stale state. `/handle` read
whatever PR export an earlier turn had left in the tree, so it could
say "nothing changed" off an hour-old snapshot; and the ladder that
routes a session's opening prompt sat resident in CLAUDE.md, where it
was routinely skipped, so a change asked for in prose got made
directly — no plan-or-not call, no quality passes, no PR.

A `UserPromptSubmit` hook now takes the PR export before the turn
starts, and `/handle` commits what it read, so the next export arrives
as a `git diff` against it: an empty diff is the answer to "has
anything changed?", not the absence of one, and the committed sequence
shows a reviewer which state of the PR each turn was answering. A
second hook puts the routing call in front of the first prompt,
pointing at the ladder rather than copying its rows. Both share one
`first_prompt` guard, and `/finalize` sweeps the export tree before
the PR goes ready so the adds and the delete cancel out here.

Three skill sections that fire on a single invocation form move to
colocated pages, paid for only when their trigger comes up — native
plan-mode recovery, the working-file pick `/squash-message` needs on a
first run, and frame extraction for a video attachment. The catalog
check gains the reverse of its reference assertion, a page nothing
points at failing as silently as a pointer to nothing. An export is
now one file however long the PR, the hoist into `threads/` having
turned one arriving comment into a ten-file commit. And `/pr`
refreshes over REST: `gh pr edit` asks for project cards on every
edit, so where classic projects are deprecated it fails and leaves the
PR exactly as it was.

The repo we vendor this loop from moved; the watermark records how far.

```

---

vzakharov:
### Attestation

Verified 3e7a50a (`claude/update-muthur-ku9drx`) against `main` @ bdf05c3 "fix: #47 the CV PDFs' dead links, and the churn a re-render left (pr #49)".

- CI-only buckets — **not dispatched: `/test-on-gh` is declined here, and the diff reaches no browser, service or credential**

Two things about what that covers, so the line above is not read for more than it says.

The vet run was green on 68e6643, the base merge. 3e7a50a is that tree minus `docs/remove-before-merging/`, and `check-squash-message.sh` passes quietly on a branch with no proposal — so nothing in the roster reads what the sweep took out.

The roster also does not reach `scripts/gh_export/`: its suite stays at the source, per the watermark's `scripts/test_*.py` row. That half was verified by running `scripts/export-github-item.py` against PRs #1 and #49 and reading the output — one `pr.md` each, no `threads/`, every index anchor resolving inside the same file. Both new hooks were driven with real payloads the same way rather than read: the PR export against #49, silent on the trunk and on a branch with no PR; the routing notice firing on a first prose prompt and staying quiet for a `/`-prefixed one and for later prompts.

---

PR #55, opened 2026-09-17
docs: the notes file stops taxing every session that touches it

vzakharov:
## Summary

- **`scripts/check-notes-length.sh` now carries two numbers, not one.** It still fails past `MAX_LINES=400`, and the failure asks for `TARGET_LINES=300`. With a single number every squeeze stopped one line under the ceiling, the next append tripped the check again, and each session that touched the file paid a crumb cut; the hundred lines between the two turn that into one editorial pass.
- **The target is asked for, not enforced** — same reason nothing in the script trims. Where to stop is a judgement; only the ceiling is a rule.
- **`writing/notes/the-five-percent.md` states the rule in its own "How this file is kept"**, which is where an agent appending to it reads, and gains a line saying a dropped bump is recoverable from history rather than lost.
- **The file then gets the squeeze the new rule demands**, 406 lines down to 296. Prose first; then near-twins merged (the `content:` prefix and the day-old word it became are one bump now, as are both readings of our own `boundaries` config); then the over-evidenced sections thinned — the top learning ran to fourteen specimens for a claim its heading already makes at twenty. Two learnings go out whole from the bottom, both at two bumps: "It settles a constraint in prose where a mechanism was available" and "A published number is not a verified number". Seven learnings survive with their counts unchanged.
- **The squeeze order itself gained a step.** It used to run tighten → cut archaeology → drop from the bottom, which assumed the bottom was coincidences with a paragraph attached. It no longer is, so merging twins and thinning a fat section now come first.
- **The editing itself moves off the session: `CLAUDE.md` hands every edit to that file to a subagent** — the review entry and any squeeze the check asks for. The session states the learning in its own words and the subagent places it, squeezes if the ceiling trips and returns when the check passes. The clauses on how to place an entry go with it: the notes file's own "How this file is kept" is the brief, so the prompt restates none of it and the session neither reads the file nor re-checks what comes back.
- **Two reasons stand behind that, and the second is the stronger one.** The first is cost, and it is the round-trip loop rather than the one edit: cut, re-measure, cut again, with placement judged against a file kept as abstraction. The second is bias — an agent that reads the taxonomy before naming its own failure names it in the headings it just read, so the count that grows is the one easiest to file under and the measurement is the thing the file exists for. Handing over the learning in the session's own words, written before the file is opened, is what removes it.
- **The criterion for what counts gained the line it was missing: a human has to be in it.** A mistake the agent caught by itself fits "something the agent could have seen and didn't" exactly and still demonstrates the opposite of what the file claims, which is that the human five percent is necessary.
- **Three bumps from this branch's own review rounds**, and a placement audit over all of them: "filed twice where the path says it gets deleted" moves to "It writes its reasoning into the artifact", whose thesis broadens to name content outliving the file it was put in (×22→×21, ×7→×8), and "Asked to quote a source" becomes "Asked for a source, it supplies its own version" — two of its specimens had no source open at all. The file stands at 318 lines, inside the ceiling and above the floor, which is the state the new rule is for.

## QA Checklist

- [ ] `pass` — `sh scripts/check-notes-length.sh` on this branch reports ok and exits 0.
- [ ] `fail` — `seq 1 455 > tmp/fake.md && sh scripts/check-notes-length.sh tmp/fake.md` exits 1 and the message names 300, not 400, as the landing point.
- [ ] `usage` — `sh scripts/check-notes-length.sh --help` states both numbers.
- [ ] `vet` — `./scripts/vet.sh` is green, the notes check included.
- [ ] `read` — read `writing/notes/the-five-percent.md` end to end: the seven surviving learnings still read as distinct, the merged bumps still carry their operator quotes, and nothing cross-references a section that went.
- [ ] `cite` — `grep` the tree for the renamed heading: `src/pages/lsa-home/ui/about-section.tsx` and `writing/late-stage-agentic/plan.md` cite "What it was handed, it treats as fixed", which is unchanged, and nothing cites the quote-source heading by its old wording.
- [ ] `recover` — `git log -p writing/notes/the-five-percent.md` shows the dropped bumps, which is what the new bullet promises.
- [ ] `delegate` — the next review session that owes an entry spawns a subagent for it, hands over the learning and nothing else, and lands the entry without the parent reading the notes file.

| `pass` | integration | ❌ | No test covers this script; a harness would run it over a fixture tree under the OS temp dir, as `type-overlap-check.test.ts` does |
| `fail` | integration | ❌ | Same harness — assert exit 1 and that the message names `TARGET_LINES` |
| `usage` | integration | ❌ | Same harness — assert `--help` prints both constants |
| `vet` | e2e | ✅ | `scripts/vet.sh` calls the script; that is the only automation it has today |
| `read` | manual-only | — | Whether seven learnings still read as distinct after a 27% cut is editorial judgement |
| `cite` | unit | ❌ | A link checker over headings this repo cites would cover it; nothing does today, so the grep was run by hand |
| `recover` | manual-only | — | Reassurance about history, not a behaviour to assert |
| `delegate` | manual-only | — | Whether an agent follows a prose rule is only observable in a later session; no check can assert it |

https://claude.ai/code/session_01VfHMiLFsHYnEwfhohij3hW

---


vzakharov:
Proposed squash title/body:

```
docs: the notes file stops taxing every session that touches it (pr #55)
```

```
`writing/notes/the-five-percent.md` charged every session that touched
it. It lived at its own 400-line ceiling and the check enforcing it
named that one number, so each squeeze stopped a line under it and the
next append tripped the check again -- a crumb cut before anything
could be added, session after session.

`scripts/check-notes-length.sh` now carries a floor as well: it fails
past 400 and the failure asks for 300, and the hundred lines between
them are what makes a squeeze last. The target is asked for rather than
enforced, for the reason nothing in the script trims -- where to stop is
a judgement, and only the ceiling is a rule. The file then takes the
squeeze the rule demands, 406 lines to 296: near-twins merged, the
over-evidenced sections thinned and the two bottom learnings dropped
whole, with the seven survivors' counts unchanged -- a count is the
claim, and the lines under it only how many of the times got written
down. The squeeze order gained two steps in the same pass, having gone
straight from tightening to dropping on the assumption that the bottom
of the file was coincidences with a paragraph attached.

The editing itself is a subagent's now, stated in `CLAUDE.md` where the
entry is mandated. Cost is the lesser reason -- the round-trip loop
rather than the one edit, cut and re-measure and cut again. The reason
is bias: an agent that reads the taxonomy before naming its own failure
names it in the headings it just read, so the count that grows is the
one easiest to file under and the measurement is what the file is for.
The session hands over the learning in its own words, written before
the file is opened, and takes back one line saying where it landed.

What counts as a bump gained the line it was missing -- a human has to
be in it, a mistake the agent caught by itself demonstrating the
opposite of the file's claim. A pass over where the existing bumps sit
re-homes one under the learning about reasoning written into the
artifact, and widens the quote-a-source heading to cover the half of
its evidence that never had a source open.

```

---

vzakharov:
### Attestation

Verified 1497154 (`claude/notes-ceiling-floor-c13f7u`) against `main` @ 0fcb36e "chore: a PR export per turn, and the entry ladder at the prompt (pr #54)".

- CI-only buckets — **not dispatched: `/test-on-gh` is not part of this project's skill set, and the diff reaches no surface the vet run misses**

---

PR #58, opened 2026-09-17
docs: read a PR's conversation from the export, not gh JSON

vzakharov:
## Summary

- **`/handle` Step 2 requires a fresh PR export, and the two steps before it fetched the same conversation through `gh` JSON first** — `/from-branch` Step 1 "before doing anything else", and Step 2's own recency signal. An agent that arrives at the export already holding the body, the reviews and the comments has no reason left to open it, so the read that picks the lane gets made from JSON with the thread structure, the resolved state and the `(agent)` / `(human)` labels flattened out of it — the labels the tail test runs on.
- **Both fetches now ask only for what the export does not carry**: `state`, `isDraft`, `baseRefName`, `mergeable`, `mergeStateStatus`. The recency test reads its dates off the export and `git log -1 --format=%cI`, and `/from-branch` states once where a PR's body and threads come from, its deep-link bullet pointing at that rather than restating it.
- **`/from-branch` Step 3 now reports what the trunk has under `.claude/` that the branch does not.** Skills, rules and hooks are read from the working tree, so a branch cut before a loop change runs the loop as it stood then — and a rule the branch's copy lacks is one the session will not follow, with nothing visible from outside to say why. This is how a `/handle` on a long-lived branch kept running a version of the skill that had already been fixed on `main`.

## QA Checklist

- [ ] `narrowed-fetch` — read `/from-branch` Step 1: the `gh pr view` field list carries no `body`, `reviews` or `comments`, and the paragraph under it names the export as where those come from
- [ ] `recency-signal` — read `/handle` Step 2's second signal: it dates comments off the export and the head off `git log`, with no `gh` call left in it
- [ ] `stale-loop` — on a branch cut before a `.claude/` change (`claude/lsa-bible-obbcfw` is one), run `git diff --stat HEAD...origin/main -- .claude CLAUDE.md` and confirm it names the files the branch is missing
- [ ] `catalog` — `scripts/check-skill-catalog.sh` passes, so every `@`-reference the new prose adds resolves
- [ ] `live-run` — point a fresh session at a PR with `/handle` and watch it pick the lane off `docs/pr/<n>/pr.md`; the failure this fixes looks like an agent answering from a `gh pr view` dump instead

| `narrowed-fetch` | manual-only | — | Agent instructions; correctness is whether a reader follows them |
| `recency-signal` | manual-only | — | Same |
| `stale-loop` | manual-only | — | The command is mechanical; that it belongs in the attach report is judgement |
| `catalog` | unit | ✅ | `scripts/check-skill-catalog.sh`, which `vet.sh` runs |
| `live-run` | manual-only | — | The behaviour under test is an agent's, observed over a real session |

https://claude.ai/code/session_01S4mmZqcrrkTkFp51PuYJ3s


vzakharov:
Proposed squash title/body:

```
docs: read a PR's conversation from the export, not gh JSON (pr #58)
```

```
`/handle` picks its lane off whether the PR carries feedback nobody has
answered, and that read needs the thread structure, the resolved state,
and the `(agent)` / `(human)` labels that separate an agent's own reply
from an operator's guidance. Only the PR export carries them. Both steps
that run before it fetched the same conversation as `gh` JSON first,
where all three are flattened away — so an agent reached the export
already holding an answer, and had no reason left to open it.

The two fetches now ask only for what the export does not carry: state,
draft flag, base ref and merge state. The recency half of the lane test
takes its dates from the export and `git log`, and `/from-branch` states
once where a PR's body and threads come from, its deep-link bullet
pointing at that rather than repeating it.

`/from-branch` also reports, on attach, what the trunk has under
`.claude/` that the branch does not. Skills, rules and hooks are read
from the working tree, so a branch cut before a loop change runs the
loop as it stood then, and nothing about running it feels any different
— a rule the branch's copy lacks is one that session will not follow,
with nothing from outside to show the operator why.

```

---

vzakharov:
Closing unmerged: the diagnosis behind it was made against a stale copy of the skills.

The session that opened this was attached to a branch cut before 0fcb36e, so it was reading `/handle` and `/from-branch` as they stood before the export hook landed. That alone explains the behaviour the operator reported. The additional claim here — that the `gh pr view --json body,reviews,comments` fetches in `/from-branch` Step 1 and `/handle` Step 2 pull an agent away from the export — was never observed failing; the session that argued it exported and read the export as the skill asks, stale copy and all.

So the change is a hypothesis dressed as a fix. Re-testing first: a fresh `/handle` on a branch carrying current `main`, to see whether anything reproduces at all. If it does, the branch `claude/handle-pr-export-obbcfw` is still pushed and this PR can be reopened over it.

---

PR #70, opened 2026-09-17
docs: plan the CV as pre-rendered markdown

vzakharov:
## Summary

- Plans moving the CV's copy out of the `cv` slice of `src/shared/i18n/messages/{en,ru}.json` and into markdown under `apps/vova/public/cv/`, compiled by `shared/content`'s build-time pipeline the way the case studies and the Bible are — so the CV is authored as prose, served raw at its route plus `.md`, and still looks exactly like the CV it looks like today.
- **Settles the file layout**: four standalone documents at `cv/<variant>/<locale>.md`, beside the `.pdf` those routes already serve. Variant is a directory rather than a dotted suffix, so the change does not touch the `<slug>.<variant>` / `<slug>.<locale>` collision `.claude/rules/content.md` parks as unsettled.
- **Settles the renderer**: compile and inject as `ArticleBody` does, with `cv.module.scss` becoming a structural stylesheet beside `prose.scss` — a CSS module's hashed class names are unreachable from injected markup. One new `:::entries` directive marks the one section whose cards are per-entry.
- **Names the consequences that reach outside `src/pages/cv`**: `scripts/lib/cv-card.ts` reads the markdown instead of the catalogue, `CV_SOURCES` hashes the `.md` instead of the catalogue slice, and all four PDFs and both social cards are re-rendered and committed.
- **Names the cost being accepted**: `cvMessages`' variant merge gives every shared sentence one home per language; four flat files give it two. The DRY notes argue why the alternative — generating the per-variant `.md` — costs the property the change exists for.

## QA Checklist

- [ ] `layout` — read the file-layout section: does variant-as-directory read right against the case studies' dotted cuts, and is four-files-with-duplication the trade you want (open question 1)?
- [ ] `renderer` — read the renderer section and "Known fidelity risks": injecting HTML means re-deriving 266 lines of `cv.module.scss`, half of it print rules keyed to per-slot classes, against structural selectors. Open question 2 offers the safer-on-fidelity alternative.
- [ ] `fidelity` — decide open question 3: may the label/lead bullet forms collapse into one authored bold, and Education's larger subheading into the standard one?
- [ ] `markdown-sample` — read the sample markdown block: is that how you want to author the CV?
- [ ] `dry-notes` — check the DRY notes' call not to register the CV in `COLLECTIONS`, and the call to inline the four `TECH_STACKS` strings rather than resolve them through a directive.

| `layout` | manual-only | — | A scope decision; nothing to assert |
| `renderer` | manual-only | — | Ditto, and the fidelity risk it carries is judged by eye |
| `fidelity` | manual-only | — | Ditto |
| `markdown-sample` | manual-only | — | Authoring ergonomics, judged by reading it |
| `dry-notes` | manual-only | — | A reuse-vs-duplication argument to agree or disagree with |

https://claude.ai/code/session_01M51DcThGuQ3RP3b4vZtZ64


vzakharov:
Proposed squash title/body:

```
feat: the CV as pre-rendered markdown (pr #70)
```

```
The CV's copy lived as a 277-line slice of the message catalogues, where
a paragraph was a JSON string and a bullet a `{label, text}` record — a
second content system beside the pipeline that already compiles the case
studies and the Bible. It goes through that pipeline now: four markdown
documents under `apps/vova/public/cv/`, compiled at build time, and
served raw at each route plus `.md` beside the PDF that used to be the
only file a reader could take away.

Variant is a directory and locale a file — `cv/cto/en.md`, beside the
`cv/cto/en.pdf` that route already served — so the CV keeps the
file-at-its-route rule without touching the dotted-suffix collision
between a document's cut and the unbuilt locale seam. Each of the four
states its framing in full: the variant merge that gave every shared
sentence one home per language is gone, and with it the guarantee that a
key `cto` left unstated could not drift from `dev`.

The sheet looks as it did, on screen and on paper. Its body compiles to
HTML and is injected as an article's is, so `cv.module.scss` becomes a
structural stylesheet beside `prose.scss` — a CSS module's hashed names
being unreachable from injected markup — and one `:::entries` directive
marks the section whose cards are per-entry. The social-card generator
and the PDF source hashes read the markdown where they read the
catalogue, and all four PDFs and both cards are re-rendered here.

```

---

